PR #2, opened 2024-07-05
SS-7800: interaction widget

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-7800

implementation of the interaction widget, which already works quite well and has quite some features.

There is currently only one known bug where the multi selection causes some materials still to be shown as selected after a deselection. I'm not sure if this is viewer side or app builder side, but will fix this next week. Just so you are aware.

We can discuss implementation details in our meeting next week.

### DIFF
--- a/components/shapediver/appbuilder/AppBuilderSelectionWidgetComponent.tsx
+++ b/components/shapediver/appbuilder/AppBuilderSelectionWidgetComponent.tsx
@@ -44,14 +44,22 @@ interface Props extends IAppBuilderWidgetPropsInteraction {
  * @param nodes 
  * @returns 
  */
-const createResponseObject = (pattern?: { [key: string]: string }, nodes?: ITreeNode[]): string => {
+const createResponseObject = (patterns?: { [key: string]: string[][] }, nodes?: ITreeNode[]): string => {
 	const getOutputAndPathFromNode = (node: ITreeNode): string | undefined => {
 		let tempNode = node;
 		while(tempNode && tempNode.parent) {
-			const outputApiData = tempNode.data.find((data) => data instanceof OutputApiData);
+			const outputApiData = tempNode.data.find((data) => data instanceof OutputApiData) as OutputApiData | undefined;
 			if(outputApiData) {
-				// const p = pattern?.[outputApiData.id];
-				return node.getPath();
+				const path = node.getPath().replace(tempNode.getPath(), "");
+				const p = patterns?.[outputApiData.api.name];
+
+				if(p) {
+					for(const pattern of p) {
+						const match = path.match(pattern.join(".*"));
+						if(match) 
+							return match[0];
+					}
+				}
 			}
 			tempNode = tempNode.parent;
 		}
@@ -71,24 +79,23 @@ const createResponseObject = (pattern?: { [key: string]: string }, nodes?: ITree
 };
 
 /**
- * Widget component for the interactions.
+ * Widget component for the selections.
  * 
  * @param param0 
  * @returns 
  */
-export default function AppBuilderInteractionWidgetComponent({ interactionSettings, parameter: p, sessionId, viewportId }: Props) {
+export default function AppBuilderSelectionWidgetComponent({ interactionSettings, parameter: p, sessionId, viewportId }: Props) {
 	// generate a unique id for the widget
 	const uuid = useId();
-
 	const settings = isInteractionSelectionParameterDefinition(interactionSettings) ? interactionSettings : undefined;
 
-	// state for the interaction application
-	const [interactionActive, setInteractionActive] = useState<boolean>(false);
-
+	// state for the selection application
+	const [selectionActive, setSelectionActive] = useState<boolean>(false);
+	// state for the selected nodes
 	const [selectedNodes, setSelectedNodes] = useState<ITreeNode[]>([]);
 
 	// get the parameter API
-	const parameter = useParameterStateless<string>(sessionId, p?.name || "");
+	const parameter = useParameterStateless<string>(sessionId, p?.name || "selected");
 	const parameterRef = useRef(parameter);
 
 	// update the parameter reference when the parameter changes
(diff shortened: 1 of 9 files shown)


types/shapediver/appbuilder.ts
```diff
@@ -84,6 +84,38 @@ export interface IAppBuilderWidgetPropsImage extends IAppBuilderWidgetPropsAncho
...
+			nodes?: {
+				output: string,
+				pattern?: string,
+				nodeName: string,
+				path: string
+			}[]
+		}
+	},
+	/** Name of the parameter that should be used. */
+	parameterName?: string
```
snabela, line 116: @MajorMeerkatThe3rd I suggest to rename this property to `parameter` and change it from `string` to `Partial<IAppBuilderParameterRef, "name" | "sessionId">`

types/shapediver/appbuildertypecheck.ts
```diff
@@ -135,6 +135,12 @@ const IAppBuilderWidgetPropsBarChartSchema = z.object({
 	style: z.enum([ "default", "stacked", "percent", "waterfall"]).optional(),
 }).extend(IAppBuilderWidgetPropsChartCommonSchema.shape);
 
+// Zod type definition for IAppBuilderWidgetPropsInteraction
+const IAppBuilderWidgetPropsInteractionSchema = z.object({
+	parameterName: z.string().optional(),
+	interactionSettings: z.any(),
```
snabela, line 141: @MajorMeerkatThe3rd once the schema is final, let's type it here as well

hooks/shapediver/viewer/useInteraction.ts
```diff
@@ -0,0 +1,159 @@
...
+			path: string
+		},
+		nodes?: {
+			output: string,
+			pattern?: string,
+			nodeName: string,
+			path: string
+		}[]
+	}
+}): {
```
snabela, line 36: @MajorMeerkatThe3rd I suggest to type the settings as an interface, export it, and use it for the App Builder widget as well

types/shapediver/appbuilderinteractiontypes.ts
```diff
@@ -0,0 +1,106 @@
...
+ * 
+ * TODO: How to define the transformation values? Should this be a matrix?
+ */
+export interface ITransformationInteractionProperties extends IGeneralInteractionProperties {
+    /** The translation values. */
+    translation?: number[][], // TODO: How to define this?
+    /** The rotation values. */
+    rotation?: number[][], // TODO: How to define this?
+    /** The scaling values. */
+    scaling?: number[][], // TODO: How to define this?
```
snabela, line 92: @MajorMeerkatThe3rd here we need to use a single property for storing the 4x4 transformation matrix. I am pretty sure one can't get separate values for translation/rotation/scaling from the Gumball, because these transformations are not commutative (the result depends on the order). Let's use the same property name and order of storing the matrix coefficients as defined by glTF.

types/shapediver/appbuilderinteractiontypes.ts
```diff
@@ -0,0 +1,106 @@
...
+ * This is a placeholder for now.
+ */
+export interface IInteractionParameterPropsDragging extends IGeneralInteractionParameterProperties {}
+
+/**
+ * Properties of a gumball parameter.
+ */
+export interface IInteractionParameterPropsGumball extends IGeneralInteractionParameterProperties {
+    /** The initial frame of the gumball. */
+    initialFrame?: any, // TODO: How to define this?
```
snabela, line 42: @MajorMeerkatThe3rd it would be sufficient to use an array of two 3-dimensional vectors. These vectors are expected to be orthonormal. Their cross-product defines the third axis.

types/shapediver/appbuilderinteractiontypes.ts
mathieu-shapediver: @snabela @MajorMeerkatThe3rd Agreed that these have to be matrices if we want to use the same objects for all types of inputs. However, it will be then likely be needed to do some sanity check / project of the matrix in case we know it is a pure transformation / rotation / scaling.

types/shapediver/appbuilderinteractiontypes.ts
mathieu-shapediver: default [[1,0,0],[0,1,0]]

types/shapediver/appbuilderinteractiontypes.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Properties of a dragging parameter.
+ * 
+ * This is a placeholder for now.
+ */
+export interface IInteractionParameterPropsDragging extends IGeneralInteractionParameterProperties {}
+
+/**
+ * Properties of a gumball parameter.
+ */
+export interface IInteractionParameterPropsGumball extends IGeneralInteractionParameterProperties {
```
mathieu-shapediver, line 40: @MajorMeerkatThe3rd @snabela We also need a minimumSelection and maximumSelection here. Would be nice to have it in the parent interface but we agreed for now that the dragging input would only allow one element at a time.

types/shapediver/appbuilderinteractiontypes.ts
snabela: @mathieu-shapediver there is a possibility to get Grasshopper to show this information, I checked it

types/shapediver/appbuilderinteractiontypes.ts
MajorMeerkatThe3rd: @mathieu-shapediver Makes sense, I forgot about that. Will be added within the next iteration.

types/shapediver/appbuilderinteractiontypes.ts
MajorMeerkatThe3rd: @snabela @mathieu-shapediver ok great, I'll change it to a single matrix within the next iteration. Much easier either way.

types/shapediver/appbuilderinteractiontypes.ts
snabela: @MajorMeerkatThe3rd are you referring to the initial scaling factors of the gumball control? I would always keep this set to `1`. The vectors defining the initial frame can be assumed to be unitized. @mathieu-shapediver do you agree?

MajorMeerkatThe3rd: @snabela @mathieu-shapediver added two new properties instead of the initial frame, `initialVectorU` and `initialVectorV`. We have a vectorU and vectorV also for the drawing tools plane definition, so I like the consistency.

components/shapediver/appbuilder/AppBuilderSelectionWidgetComponent.tsx
MajorMeerkatThe3rd, line 1: @mathieu-shapediver This fixes the filtering issues we discussed.

There was just no start and end to my regex check, so everything before and after was ignored.

mathieu-shapediver (approved):
looks good!

PR #3, opened 2024-07-11
SS-7607 platform/model store refactoring

snabela:
https://shapediver.atlassian.net/browse/SS-7607


snabela:
@LemToUp @mathieu-shapediver I will merge this right away, but it's anyway interesting for you to review this. The functionality in this PR used by the [PERI-specific App Builder](https://github.com/shapediver/AppBuilderPERI). The same shared functionality will be used for the main App Builder once this PR is merged. Some notable items: 

  * This file shows the configuration of library tabs for PERI: https://github.com/shapediver/AppBuilderPERI/blob/development/src/hooks/useLibraryTabConfig.ts
  * The library tabs also work if you are logged in using a non-PERI platform user, however the configuration of tabs is different if the user is not member of an organization. Give it a try [here](https://www.shapediver.com/app/builder/v1/peri/latest/)
  * I implemented a generic store for platform items, and its implementation for models. This works really well already, as you can try on the library page I linked above. Query results are cached, and the cache gets pruned if necessary (e.g. if bookmarks are changed, or the confirmation status of a model). Related code: https://github.com/shapediver/AppBuilderShared/pull/3/files#diff-bfa1e5816303da86543eee690553b81fa2d21e0845c6a7b459217aeb898f6307
  * The model store is used by the [ModelLibrary](https://github.com/shapediver/AppBuilderShared/pull/3/files#diff-6dd4dc02b4005f0d4d81ce55e7628e846fbeccb2c5c421d0e6a42f77f5c5e895) component
  * The model card is nicely extensible: https://github.com/shapediver/AppBuilderShared/pull/3/files#diff-72e8cd5381f8d65de7591f46b08da949a5538dfc8d4aa46fbd4ebf0d28cf2f0b
  * It makes use of a new [ToogleIcon](https://github.com/shapediver/AppBuilderShared/pull/3/files#diff-076313e2d17c37cfa474a2b0a228fc9cc5b5a971ddd6e28633a1a97a3ea77acc) component for toggling the bookmark status, and the confirmation status. 
  * In the [model store implementation](https://github.com/shapediver/AppBuilderShared/pull/3/files#diff-b6375edd3cbdf4751651cd17070649d62cc29fcc5c5291ed926eb78909208fe4) you can see the "action" functions available for models already (bookmarking, confirmation). These functions update the local state of the model (such that the change appears on all instances of the model card on different tabs immediately), make the API call to update the model on the platform, and prune the cache of model queries (wherever necessary).

PR #6, opened 2024-08-07
SS-7800 SS-7851 Selection Parameter

MajorMeerkatThe3rd:
@snabela please review the last commit, there is still a small issue, but it overall seems to work well


MajorMeerkatThe3rd:
@snabela Now it's the last two commits, everything works now as expected!

hooks/shapediver/viewer/interaction/useInteractionEngine.ts
snabela, line 1: @MajorMeerkatThe3rd we only want one interaction engine per viewport, correct? If so, we need to store them in a global object defined outside of the hook, which maps viewport id to interaction engine

hooks/shapediver/viewer/interaction/useHoverManager.ts
```diff
@@ -0,0 +1,56 @@
...
+
+	// call the interaction engine hook
+	const { interactionEngine } = useInteractionEngine(viewportId);
+
+	// create a reference for the hover manager
+	const hoverManagerRef = useRef<HoverManager | undefined>(undefined);
+	// create a reference for the hover manager token
+	const hoverManagerTokenRef = useRef<string | undefined>(undefined);
+
+	// use an effect to apply changes to the material, and to apply the callback once the node is available
```
snabela, line 31: the comment is a leftover ;-) it also appears in other places

hooks/shapediver/viewer/interaction/useHoverManager.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param viewportId 
+ */
+export function useHoverManager(viewportId: string, settings?: IInteractionParameterSettings): {
+	/**
+	 * The hover manager.
+	 */
+    hoverManager?: HoverManager
+} {
+	// get the viewport API
+	const viewportApi = useShapeDiverStoreViewer(state => { return state.viewports[viewportId]; });
```
snabela, line 21: @MajorMeerkatThe3rd I think we don't need the viewportApi here, it should be sufficient to check for the interactionEngine

hooks/shapediver/viewer/interaction/useSelectManager.ts
```diff
@@ -0,0 +1,102 @@
...
+	 * The select manager, if the parameter settings only allow single selection.
+	 */
+	selectManager?: SelectManager,
+	/**
+	 * The multi select manager, if the parameter settings allow multiple selection.
+	 */
+	multiSelectManager?: MultiSelectManager
+} {
+	// get the viewport API
+	const viewportApi = useShapeDiverStoreViewer(state => { return state.viewports[viewportId]; });
```
snabela, line 25: @MajorMeerkatThe3rd probably we can get rid of the viewportApi here as well

hooks/shapediver/viewer/interaction/useHoverManagerEvents.ts
```diff
@@ -0,0 +1,66 @@
...
+import { processNodes } from "./utils/patternUtils";
+
+// #region Functions (1)
+
+/**
+ * Hook allowing to create the hover manager events.
+ * 
+ * @param viewportId 
+ */
+export function useHoverManagerEvents(pattern: { [key: string]: string[][] }): {
```
snabela, line 13: @MajorMeerkatThe3rd let's document the parameter

hooks/shapediver/viewer/interaction/useSelectManagerEvents.ts
```diff
@@ -0,0 +1,133 @@
...
+	// register an event handler and listen for output updates
+	useEffect(() => {
+		/**
+         * Event handler for the select on event.
+         * In this event handler, the response object is created and the parameter is updated.
+         */
+		const tokenSelectOn = addListener(EVENTTYPE_INTERACTION.SELECT_ON, async (event: IEvent) => {
+			const selectEvent = event as InteractionEventResponseMapping[EVENTTYPE_INTERACTION.SELECT_ON];
+
+			// don't send the customization if the event is coming from an API call
```
snabela, line 39: @MajorMeerkatThe3rd this comment and others seem a bit misleading to me, maybe better: _We ignore the event if it's not based on an event triggered by the UI._ (or similar).

hooks/shapediver/viewer/interaction/useSelectManagerEvents.ts
```diff
@@ -0,0 +1,133 @@
...
+			if (!multiSelectEvent.event) return;
+
+			const selected = multiSelectEvent.nodes;
+			const nodeNames = processNodes(pattern, selected);
+			setSelectedNodeNames(nodeNames.names);
+
+			if (multiSelectEvent.nodes.length < (multiSelectEvent.manager as MultiSelectManager).minimumNodes) return;
+			if (multiSelectEvent.nodes.length > (multiSelectEvent.manager as MultiSelectManager).maximumNodes) return;
```
snabela, line 77: @MajorMeerkatThe3rd probably this should be moved up before setSelectedNodeNames?

hooks/shapediver/viewer/interaction/useHoverManager.ts
```diff
@@ -0,0 +1,56 @@
...
+import { useInteractionEngine } from "./useInteractionEngine";
+
+// #region Functions (1)
+
+/**
+ * Hook allowing to create the hoverManager.
+ * 
+ * @param viewportId 
+ */
+export function useHoverManager(viewportId: string, settings?: IInteractionParameterSettings): {
```
snabela, line 14: @MajorMeerkatThe3rd let's not use IInteractionParameterSettings here, but the type of settings that make sense for the HoverManager. Same for other hooks.

hooks/shapediver/viewer/interaction/utils/patternUtils.ts
```diff
@@ -0,0 +1,92 @@
...
+		if (typeof settings.props.nameFilter === "string") {
+			if ((settings.props.nameFilter as string).startsWith("[") && (settings.props.nameFilter as string).endsWith("]") && (settings.props.nameFilter as string) !== "[]") {
+				try {
+					nameFilter = JSON.parse(settings.props.nameFilter) as string[];
+				} catch (e) {
+					notifications.show({
+						title: "Invalid Name Filter",
+						message: "The name filter is not a valid JSON array."
+					});
+				}
```
snabela, line 28: @MajorMeerkatThe3rd this should never happen, let's add validation where we receive the data (e.g. in appbuildertypecheck.ts)

hooks/shapediver/viewer/interaction/utils/patternUtils.ts
snabela, line 1: @MajorMeerkatThe3rd some of this appears strange to me, as does `useProcessPattern`. Let's refactor this further. I wonder whether the sessionApi is really required here.

types/shapediver/appbuildertypecheck.ts
```diff
@@ -136,6 +136,12 @@ const IAppBuilderWidgetPropsBarChartSchema = z.object({
 	style: z.enum([ "default", "stacked", "percent", "waterfall"]).optional(),
 }).extend(IAppBuilderWidgetPropsChartCommonSchema.shape);
 
+// Zod type definition for IAppBuilderWidgetPropsInteraction
+const IAppBuilderWidgetPropsInteractionSchema = z.object({
+	parameterName: z.string().optional(),
+	interactionSettings: z.any(),
```
snabela, line 142: @MajorMeerkatThe3rd let's add the complete type definition here (we can limit this to the properties we use in the widget)

hooks/shapediver/viewer/interaction/useNodeInteractionData.ts
```diff
@@ -0,0 +1,125 @@
+import { GeometryData, IOutputApi, ITreeNode } from "@shapediver/viewer";
+import { useCallback, useEffect } from "react";
+import { IInteractionData, InteractionData } from "@shapediver/viewer.features.interaction";
+import { useOutputNode } from "../useOutputNode";
+
+const nodesWithInteractionData: { [key: string]: { node: ITreeNode, data: IInteractionData } } = {};
```
snabela, line 6: @MajorMeerkatThe3rd let's document what exactly we store here (what the key values are)

MajorMeerkatThe3rd:
@snabela I just pushed another refactoring, please review when you have a sec!

hooks/shapediver/viewer/interaction/useSelectManager.ts
```diff
@@ -0,0 +1,102 @@
...
+	}
+};
+
+/**
+ * Hook allowing to create a select manager for a viewport.
+ * 
+ * @param viewportId The ID of the viewport.
+ * @param settings The settings for the select manager. If the settings are not provided, the select manager will not be created.
+ */
+export function useSelectManager(viewportId: string, settings?: ISelectionParameterSettings): {
```
snabela, line 41: @MajorMeerkatThe3rd small detail: I suggest to use `Pick< ISelectionParameterSettings, "minimumSelection" | "maximumSelection" | "selectionColor" >` here instead of the complete `ISelectionParameterSettings`

hooks/shapediver/viewer/interaction/useSelectManager.ts
```diff
@@ -0,0 +1,102 @@
...
+		if (settings) {
+			// whenever this output node changes, we want to create the interaction engine
+			const selectMultiple = (settings.minimumSelection !== undefined && settings.maximumSelection !== undefined) &&
+				settings.minimumSelection < settings.maximumSelection && settings.maximumSelection > 1;
+
+			// check if a select manager already exists for the viewport, but with different settings
+			// in this case we need to remove the old select manager and create a new one
+			if (selectManagers[viewportId] && selectManagers[viewportId].selectMultiple !== selectMultiple) {
+				cleanUpSelectManager(viewportId, interactionEngine);
+			}
```
snabela, line 62: @MajorMeerkatThe3rd there might be situations were there are two selection inputs, both activated, one of the for one object only, the other one for multiple objects. This would cause a problem. For now we can issue a warning to the console, but can we maybe use `MultiSelectManager` only?

hooks/shapediver/viewer/interaction/utils/patternUtils.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	// we iterate over the name filter array
+	// we store the result with the output ID as the key and an array of patterns as the value
+	for (let i = 0; i < nameFilter.length; i++) {
+		const parts = nameFilter[i].split(".");
+		const outputName = parts[0];
+		
+		// find the output Ids that match the output name
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		const outputIds = Object.entries(outputIdsToNamesMapping).filter(([id, name]) => name === outputName).map(([outputId, _]) => outputId);
```
snabela, line 49: Small optimization possibility: Outside of the loop we could create an object holding the inverse mapping of `outputIdsToNamesMapping`

snabela: I just realized that this doesn't support wildcards for the output name (the first component of the name filter). Is this a bug or a feature? @MajorMeerkatThe3rd @mathieu-shapediver

hooks/shapediver/viewer/interaction/useSelectManagerEvents.ts
```diff
@@ -0,0 +1,133 @@
...
+} {
+	// state for the selected nodes
+	const [selectedNodeNames, setSelectedNodeNames] = useState<string[]>([]);
+	const resetSelectedNodeNames = useCallback(() => setSelectedNodeNames([]), []);
+
+	// register an event handler and listen for output updates
+	useEffect(() => {
+		/**
+         * Event handler for the select on event.
+         * In this event handler, the response object is created and the parameter is updated.
```
snabela, line 34: @MajorMeerkatThe3rd  Cosmetics: This comment is a leftover (appears similar in other places)

hooks/shapediver/viewer/interaction/useProcessPattern.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Hook that processes a pattern for a session.
+ * 
+ * @param sessionId The ID of the session.
+ * @param nameFilter The name filter to apply to the pattern.
+ */
+export function useProcessPattern(sessionId: string, nameFilter?: string[]): {
+    pattern: {
+        [key: string]: string[][];
+    }
```
snabela, line 16: @MajorMeerkatThe3rd here you can use the type exported from `patternUtils`

hooks/shapediver/viewer/interaction/useProcessPattern.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// #region Functions (1)
+
+/**
+ * Hook that processes a pattern for a session.
+ * 
+ * @param sessionId The ID of the session.
+ * @param nameFilter The name filter to apply to the pattern.
+ */
+export function useProcessPattern(sessionId: string, nameFilter?: string[]): {
```
snabela, line 13: @MajorMeerkatThe3rd I suggest renaming this to sth like `useCreateNameFilterPattern`

hooks/shapediver/viewer/interaction/useProcessPattern.ts
```diff
@@ -0,0 +1,37 @@
...
+	// create a state for the pattern
+	const [pattern, setPattern] = useState<{ [key: string]: string[][]; }>({});
+
+	useEffect(() => {
+		if (nameFilter !== undefined) {
+			const outputIdsToNamesMapping: { [key: string]: string } = {};
+			Object.entries(sessionApi.outputs).forEach(([outputId, output]) => outputIdsToNamesMapping[outputId] = output.name);
+			setPattern(processPattern(nameFilter, outputIdsToNamesMapping));
+		}
+	}, [nameFilter]);
```
snabela, line 30: @MajorMeerkatThe3rd to be sure let's also use `sessionApi` as a dependency

hooks/shapediver/viewer/interaction/useSelection.ts
```diff
@@ -0,0 +1,100 @@
...
+					const parts = nodeName.split(".");
+
+					// check if the node name matches the pattern
+					const outputApi = node.data.find(d => d instanceof OutputApiData) as OutputApiData;
+					if (!outputApi) return;
+					if (outputApi.api.name !== parts[0]) return;
+
+					// if the node name matches the pattern, select the node
+					node.traverse(n => {
+						if (n.getPath().endsWith(parts.slice(1).join("."))) {
```
snabela, line 76: @MajorMeerkatThe3rd I think here the regex match is missing

hooks/shapediver/viewer/interaction/useSelection.ts
```diff
@@ -0,0 +1,100 @@
...
+			}
+		}
+	}, [selectedNodeNames, selectManager, settings]);
+
+	// add interaction data for each output, even if it is not in the pattern
+	// this is necessary to keep the number of hooks constant
+	for (const outputId in sessionApi.outputs) {
+		if (!pattern[outputId]) pattern[outputId] = [];
+		useNodeInteractionData(sessionId, outputId, pattern[outputId], { select: !!selectionSettings, hover: settings?.props.hover}, callback);
+	}
```
snabela, line 92: @MajorMeerkatThe3rd as a general comment: In the future we could avoid this loop if we implement a `useOutputUpdateCallback` hook that works with all outputs of a session. And a corresponding `useOutputNodes` hook etc.

hooks/shapediver/viewer/interaction/useNodeInteractionData.ts
```diff
@@ -0,0 +1,128 @@
...
+			for(const id in nodesWithInteractionData) {
+				const { node, data } = nodesWithInteractionData[id];
+				node.removeData(data);
+				node.updateVersion();
+			}
+			// clear the dictionary
+			Object.keys(nodesWithInteractionData).forEach(key => delete nodesWithInteractionData[key]);
+		};
+		
+	}, [interactionSettings, patterns, additionalUpdateCallback]);
```
snabela, line 122: @MajorMeerkatThe3rd let's also add `callback` to the dependencies (although it's covered by the other dependencies, but who knows what we change in the future)

hooks/shapediver/viewer/interaction/useSelection.ts
```diff
@@ -0,0 +1,100 @@
...
+				});
+			}
+		}
+	}, [selectedNodeNames, selectManager, settings]);
+
+	// add interaction data for each output, even if it is not in the pattern
+	// this is necessary to keep the number of hooks constant
+	for (const outputId in sessionApi.outputs) {
+		if (!pattern[outputId]) pattern[outputId] = [];
+		useNodeInteractionData(sessionId, outputId, pattern[outputId], { select: !!selectionSettings, hover: settings?.props.hover}, callback);
```
snabela, line 91: @MajorMeerkatThe3rd here is a trap one easily falls into: creating the settings object on the fly here results in a new instance at every re-rendering! It's important that we create this object outside of the loop using a `useMemo` hook.

components/shapediver/appbuilder/AppBuilderSelectionWidgetComponent.tsx
snabela, line 1: @MajorMeerkatThe3rd let's leave a comment here that this widget will most likely be refactored once we implement the App Builder selection widget component in Grasshopper.

snabela (commented):
@MajorMeerkatThe3rd great work! I left some comments, but just small stuff.

hooks/shapediver/viewer/interaction/useSelectManager.ts
MajorMeerkatThe3rd: @snabela For now that would be a general problem. As there is no assignment for a selection engine to the node that it can intersect it means that any node with an interaction data can be selected by any selection manager. This has to be implemented on the viewer side first.

hooks/shapediver/viewer/interaction/utils/patternUtils.ts
MajorMeerkatThe3rd: @snabela I was not aware that this was even an option, so from my side, the implementation is correct.

hooks/shapediver/viewer/interaction/useSelection.ts
MajorMeerkatThe3rd: @snabela In this case I compare the actual selected node names with the path, here it should be an exact match, so the "endsWith" is correct

hooks/shapediver/viewer/interaction/useSelection.ts
MajorMeerkatThe3rd: @snabela Makes sense, let's talk after the vacation in the office on how we could do that, I don't like the loop either :D

hooks/shapediver/viewer/interaction/useSelectManager.ts
snabela: @MajorMeerkatThe3rd got it. We can test and fix this as an extension. Probably we will use some global state to check whether another selection is already ongoing, and block activating a second one.

hooks/shapediver/viewer/interaction/utils/patternUtils.ts
snabela: @mathieu-shapediver let's make sure to document it like this. We might add additional validation for the name filters in Grasshopper (no wildcards allowed in the first component).

hooks/shapediver/viewer/interaction/useSelectManager.ts
snabela: https://shapediver.atlassian.net/browse/SS-7920

snabela (approved):

snabela:
@MajorMeerkatThe3rd I pushed two more commits: 
https://github.com/shapediver/AppBuilderShared/pull/6/commits/f6db19b6e8e443c99062ae20692a1427d984ec9e
https://github.com/shapediver/AppBuilderShared/pull/6/commits/6198f8058f705d3ef7f3afdda30e16279a57ce03

(it should be possible to configure the autoformatter to do the path normalization automatically, but not sure how)

components/shapediver/parameter/ParameterSelectionComponent.tsx
```diff
@@ -0,0 +1,110 @@
...
+				<Button justify="space-between" fullWidth h="100%" disabled={disabled}
+					rightSection={selectionActive ? iconInfo : iconFinger}
+					bg={selectionActive ? "" : "blue"}
+					onClick={() => setSelectionActive(!selectionActive)}>
+					<Group justify="space-between" w="100%" pt={"sm"} pb={"sm"}>
+						{selectionActive ? contentActive : contentInactive}
+					</Group>
+				</Button>
+			}
+		</Group>
```
snabela, line 107: @MajorMeerkatThe3rd a general comment about using styling properties: Let's keep them as minimal as possible, for example try to avoid font weight, explicit definition of padding, etc. Keeping this minimal ensures there will be no problems with the theme (we want things to look good even if someone changes padding globally via the theme). 
In case specifying specific styles can't be avoided, there is also the possibility to get theme properties (like default padding) and use them, or define theme properties for the component. There are lots of examples on how to do that in other components. 
@mathieu-shapediver FYI

PR #10, opened 2024-08-22
SS-7800 styling improvements, refactoring

snabela:


snabela:
@MajorMeerkatThe3rd this ended up becoming a quite massive refactoring. While working on it I noticed some problems to be fixed, and also significant unnecessary re-rendering due to lots of state updates. I am pretty happy with the state now. For sure there will be conflicts with the gumball PR. I will add some comments for you to this PR until Monday.

snabela (commented):
@MajorMeerkatThe3rd it might be best if you check out the code and see the changes in the IDE.

MajorMeerkatThe3rd:
@snabela I pushed two commits, please review.

The first one should be pretty clear, the second one solves the issue that I showed you this morning.

The problem was the following: When the output node changed, the interaction data was removed, but the selection manager still had the nodes marked as selected. Therefore, I needed to deselect them, which I now do in the useInteractionNodeData hook. 
This works also if there is no SelectManager present.

MajorMeerkatThe3rd:
@snabela the current viewer release is only a candidate, I'll deploy a real version tomorrow morning, then we can merge!

PR #11, opened 2024-08-23
SS-6180: refactored gumball implementation

MajorMeerkatThe3rd:
@snabela while the parameter component is not done yet as some more spec is needed, the hooks can already be reviewed!

The changes in the selection parts are actually pretty minor, I now not only return the selected node names, but additionally the selected nodes themselves. As I already had them in the selection events either way, it made sense to not traverse the scene tree another time.

### DIFF
--- a/hooks/shapediver/viewer/interaction/gumball/useGumball.ts
+++ b/hooks/shapediver/viewer/interaction/gumball/useGumball.ts
@@ -1,0 +1,75 @@
+import { useEffect, useMemo, useRef } from "react";
+import { Gumball } from "@shapediver/viewer.features.gumball";
+import { IInteractionParameterSettings, ISelectionParameterSettings, isInteractionGumballParameterSettings, ITreeNode } from "@shapediver/viewer";
+import { useShapeDiverStoreViewer } from "shared/store/useShapeDiverStoreViewer";
+import { useSelection } from "../selection/useSelection";
+import { useGumballEvents } from "./useGumballEvents";
+
+// #region Functions (1)
+
+/**
+ * Hook allowing to create the interaction engine and the managers that are specified via the settings.
+ * 
+ * @param viewportId 
+ */
+export function useGumball(sessionId: string, viewportId: string, settings?: IInteractionParameterSettings): {
+	transformedNodes: {
+		name: string,
+		transformation: number[]
+	}[],
+	setTransformedNodes: (nodes: { name: string, transformation: number[] }[]) => void,
+	selectedNodes: { name: string, node: ITreeNode }[],
+	setSelectedNodes: (selectedNodes: { name: string, node: ITreeNode }[]) => void
+} {
+	// get the viewport API
+	const viewportApi = useShapeDiverStoreViewer(state => { return state.viewports[viewportId]; });
+	// check if the settings are gumball settings
+	const gumballSettings = isInteractionGumballParameterSettings(settings) ? settings : undefined;
+	// create the selection settings from the interaction settings
+	const selectionSettings = useMemo(() => {
+		if (!gumballSettings) return undefined;
+
+		return {
+			type: "selection",
+			props: {
+				nameFilter: gumballSettings?.props.nameFilter,
+				hover: gumballSettings?.props.hover,
+				minimumSelection: 0,
+				maximumSelection: Infinity
+			} as ISelectionParameterSettings
+		} as IInteractionParameterSettings;
+	}, [gumballSettings]);
+
+	// get the selected nodes
+	const { selectedNodes, setSelectedNodes } = useSelection(sessionId, viewportId, selectionSettings);
+	// create a reference for the gumball
+	const gumballRef = useRef<Gumball | undefined>(undefined);
+
+	// use an effect to apply changes to the material, and to apply the callback once the node is available
+	useEffect(() => {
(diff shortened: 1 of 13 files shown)


hooks/shapediver/viewer/interaction/gumball/useGumballEvents.ts
```diff
@@ -0,0 +1,80 @@
...
+
+			// Set the new array
+			setTransformedNodes(newTransformedNodes);
+
+		});
+
+		return () => {
+			removeListener(token);
+		};
+	}, [selectedNodes, transformedNodes]);
```
snabela, line 72: @MajorMeerkatThe3rd let's implement this hook such that the event listener is not re-recreated every time the state of selected or transformed nodes changes.

snabela:
> @snabela while the parameter component is not done yet as some more spec is needed, the hooks can already be reviewed!
> 
> The changes in the selection parts are actually pretty minor, I now not only return the selected node names, but additionally the selected nodes themselves. As I already had them in the selection events either way, it made sense to not traverse the scene tree another time.

Looks good! I am sorry that some refactoring will be necessary, due to what I did in https://github.com/shapediver/AppBuilderShared/pull/10
🙃

MajorMeerkatThe3rd:
@snabela please review the new changes, thank you!

hooks/shapediver/viewer/interaction/gumball/useGumball.ts
```diff
@@ -0,0 +1,174 @@
...
+	// use the gumball events hook to get the transformed node names
+	const { transformedNodeNames, setTransformedNodeNames } = useGumballEvents(selectedNodeNames, initialTransformedNodeNames);
+
+	// create a reference for the gumball
+	const gumballRef = useRef<Gumball | undefined>(undefined);
+
+	// use an effect to create the gumball whenever the selected node names change
+	useEffect(() => {
+		if (viewportApi) {
+			// whenever this output node changes, we want to create the interaction engine
```
snabela, line 93: @MajorMeerkatThe3rd the comment is not clear to me, is it a leftover?

hooks/shapediver/viewer/interaction/gumball/useGumball.ts
```diff
@@ -0,0 +1,174 @@
...
+	 * Restore the transformed node names.
+	 * 
+	 * This function is used to restore the transformed nodes to their new transformation state.
+	 * This means that the transformation of the nodes is updated to the new transformation state.
+	 * 
+	 * @param newTransformedNodeNames The new transformed node names.
+	 * @param oldTransformedNodeNames The old transformed node names.
+	 * @returns
+	 */
+	const restoreTransformedNodeNames = useCallback((newTransformedNodeNames: { name: string, transformation: number[] }[], oldTransformedNodeNames: { name: string, transformation: number[] }[]) => {
```
snabela, line 118: @MajorMeerkatThe3rd detail: we could remove the transformation property from the type definition for `oldTransformedNodeNames`

hooks/shapediver/viewer/interaction/gumball/useGumballEvents.ts
```diff
@@ -0,0 +1,89 @@
...
+			const newTransformedNodeNames = [...transformedNodeNames];
+
+			for (let i = 0; i < gumballEvent.nodes.length; i++) {
+				const node = gumballEvent.nodes[i];
+				const transformation = gumballEvent.transformations[i];
+
+				// search for the node in the selected nodes
+				selectedNodeNames.forEach((name) => {
+					const parts = name.split(".");
+					if (node.getPath().endsWith(parts.slice(1).join("."))) {
```
snabela, line 54: @MajorMeerkatThe3rd do we not check for the output name (parts[0]) here on purpose?

hooks/shapediver/viewer/interaction/gumball/useGumballEvents.ts
```diff
@@ -0,0 +1,89 @@
...
+	// state for the transformed nodes
+	const [transformedNodeNames, setTransformedNodeNames] = useState<{ name: string, transformation: number[] }[]>(initialTransformedNodeNames ?? []);
+
+	// register an event handler and listen for output updates
+	useEffect(() => {
+		const token = addListener(EVENTTYPE_GUMBALL.MATRIX_CHANGED, (e) => {
+			const gumballEvent = e as GumballEventResponseMapping[EVENTTYPE_GUMBALL.MATRIX_CHANGED];
+
+			// Create a new array to avoid mutating the state directly
+			const newTransformedNodeNames = [...transformedNodeNames];
```
snabela, line 45: @MajorMeerkatThe3rd the use of `transformedNodeNames` (which is not a dependency, and ideally wouldn't be made a dependency), could lead to trouble, depending on how this hook gets used. Example: If `setTransformedNodeNames` gets called by a caller of the `useGumballEvents` hook, then the gumball event callback will still use the old instance of `transformedNodeNames`. 
Potential solution: Also define a `useRef` below the definition of `transformedNodeNames, set TransformedNodeNames`, and always update its value to `transformedNodeNames`. Use the reference in the event handler callback.

snabela (changes requested):
Many thanks @MajorMeerkatThe3rd ! I left some comments.

snabela (approved):

PR #16, opened 2024-09-03
SS-7980: activate Gumball if it is only one object

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-7980

Everything works as expected already, I'm just not entirely happy with the new `useEffect` call in the outputs loop. Do you have any better idea on how to solve that?

### DIFF
--- a/hooks/shapediver/viewer/interaction/selection/useSelection.ts
+++ b/hooks/shapediver/viewer/interaction/selection/useSelection.ts
@@ -1,6 +1,6 @@
 import { ISelectionParameterProps, ITreeNode, OutputApiData } from "@shapediver/viewer";
 import { InteractionData, MultiSelectManager, SelectManager } from "@shapediver/viewer.features.interaction";
-import { useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { vec3 } from "gl-matrix";
 import { ISelectionState, useSelectManagerEvents } from "./useSelectManagerEvents";
 import { useSelectManager } from "./useSelectManager";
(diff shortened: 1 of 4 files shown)


hooks/shapediver/viewer/interaction/selection/useSelection.ts
```diff
@@ -53,18 +61,26 @@ export function useSelection(
...
 			if (outputNode && selectManager)
 				restoreSelection(outputNode, selectManager, selectedNodeNames);
 		}, [outputNode, selectManager]);
+
+		// update the available node names
+		useEffect(() => {
+			setAvailableNodeNames(prev => {
+				return { ...prev, [outputs[outputId].name]: availableNodeNamesForOutput };
+			});
+		}, [availableNodeNamesForOutput]);
```
snabela, line 76: @MajorMeerkatThe3rd here is a potential problem: If I understood correctly, `availableNodeNames` never gets reset. I suggest this: 
   * Inside the loop, collect the data from `availableNodeNamesForOutput` into an array `nodeNames` that is not stateful (no useEffect)
   * After the loop, use an useEffect hook to compare the content of `nodeNames` with `availableNodeNames`, call `setAvailableNodeNames` if they are different. 
 
This will also minimizes the number of re-renders

snabela (changes requested):
@MajorMeerkatThe3rd many thanks, this looks good! I found a potential issue, please see the comment.

hooks/shapediver/viewer/interaction/selection/useSelection.ts
MajorMeerkatThe3rd: @snabela I tried the approach you suggest, but this ends in an endless loop. I also tried some other approaches, but they are have the same number of calls that are used. 

I think this is something that should be done in a general refactoring, of the for loop, which is probably a bigger task.

MajorMeerkatThe3rd:
@snabela I pushed another commit with a bugfix, as the selection wasn't handled properly! Please review

snabela (approved):

PR #17, opened 2024-09-05
SS-7987: parsing error

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-7987

parsing error because I missed an `s` for transformations 😂


snabela (approved):
Many thanks @MajorMeerkatThe3rd

PR #27, opened 2024-10-01
SS-7920: enabling multiple selection components at once

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-7920

@snabela As you can see, this turned out easier than expected. I just pass the ID through to the managers / interaction data so that an assignment can be achieved.


snabela (approved):
Super, this looks great @MajorMeerkatThe3rd !

PR #29, opened 2024-10-09
SS-8084 add validators for custom parameters

snabela:
https://shapediver.atlassian.net/browse/SS-8084


MajorMeerkatThe3rd (approved):
@snabela thank you!

PR #31, opened 2024-10-15
SS-8107 Implementing handling of "context" query string parameter

snabela:
https://shapediver.atlassian.net/browse/SS-8107

### DIFF
--- a/store/useShapeDiverStoreViewer.ts
+++ b/store/useShapeDiverStoreViewer.ts
@@ -1,4 +1,4 @@
-import { createSession, createViewport, ISessionApi, isViewerGeometryBackendResponseError, isViewerValidationError, IViewportApi } from "@shapediver/viewer";
+import { createSession, createViewport, ISessionApi, isViewerGeometryBackendResponseError, IViewportApi } from "@shapediver/viewer";
 import { SessionCreateDto, IShapeDiverStoreViewer, ViewportCreateDto } from "../types/store/shapediverStoreViewer";
 import { create } from "zustand";
 import { devtools } from "zustand/middleware";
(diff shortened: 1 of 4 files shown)


store/useShapeDiverStoreViewer.ts
```diff
@@ -111,7 +112,24 @@ export const useShapeDiverStoreViewer = create<IShapeDiverStoreViewer>()(devtool
...
+			}
+			catch (e: any) {
+				if (isViewerGeometryBackendResponseError(e) && 
+					e.geometryBackendErrorType === ShapeDiverResponseErrorType.REQUEST_VALIDATION_ERROR &&
+					e.message.startsWith("Invalid parameter") && e.message.includes("'context'")) 
+				{
+					console.warn("Session creation failed due to invalid or missing 'context' parameter. Retrying without 'context' parameter.");
+
+					const dtoWithoutContext: SessionCreateDto = {...dto, initialParameterValues: {...dto.initialParameterValues}};
+					delete dtoWithoutContext.initialParameterValues!["context"];
```
MajorMeerkatThe3rd, line 126: @snabela I'm not fully getting this line. If you assign the dto.initialParameterValues in the line before that, why is there a need to remove the context here? It shouldn't be there in this object anyway, right?

MajorMeerkatThe3rd (commented):
@snabela I left a comment on a line that is not completely clear to me. Besides that all good!

store/useShapeDiverStoreViewer.ts
snabela: @MajorMeerkatThe3rd the line above just deep copies `dto.initialParameterValues`, but it doesn't remove `context`. Please note that `context` is already part of the `dto` object when this function is called. Does this answer your concern?

MajorMeerkatThe3rd: @snabela Ah got it, I thought that was one level above that

MajorMeerkatThe3rd (approved):
@snabela All good, ready for merge 🚀

PR #32, opened 2024-10-16
SS-8112: reject button fix for gumball and code improvement for onCancel

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8112

@snabela small note on the difference in behavior for the gumball and selection. In the gumball, the assigned value is internally completely reset, as the transformations come from GH after a customization update.


snabela (approved):
Merci @MajorMeerkatThe3rd !

PR #33, opened 2024-10-17
SS-8123: added inner-component dirty flag

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8123

@snabela I added a dirty flag within the component to deactivate the confirm button when there is nothing to confirm


snabela (approved):
@MajorMeerkatThe3rd got it, that makes sense!

PR #34, opened 2024-10-17
SS-8124: custom selection parameters

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8124

This PR is not about actually wanting to merge this, but rather show the issue I want to discuss.

With the misuse of the sessionId by custom parameters, the sessionId cannot be used as expected. 
My idea to solve this would be to refactor how the sessionId is currently used on the custom parameters, as this doesn't make much sense to me, but I didn't want to just start with that but discuss it first.

### DIFF
--- a/components/shapediver/parameter/ParameterSelectionComponent.tsx
+++ b/components/shapediver/parameter/ParameterSelectionComponent.tsx
@@ -39,6 +39,7 @@ export default function ParameterSelectionComponent(props: PropsParameter) {
 		onCancel,
 		disabled,
 		value,
+		sessionDependencies,
 		state
 	} = useParameterComponentCommons<string>(props);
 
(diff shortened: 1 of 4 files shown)


snabela:
> https://shapediver.atlassian.net/browse/SS-8124
> 
> This PR is not about actually wanting to merge this, but rather show the issue I want to discuss.
> 
> With the misuse of the sessionId by custom parameters, the sessionId cannot be used as expected. My idea to solve this would be to refactor how the sessionId is currently used on the custom parameters, as this doesn't make much sense to me, but I didn't want to just start with that but discuss it first.

Many thanks @MajorMeerkatThe3rd, I see. Let's discuss this after our dev meeting on Friday.

PR #35, opened 2024-10-18
SS-8129: renaming of sessionId an usage of sessionDependencies

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8129

@snabela Please also search for the sessionId properties that are still present. I think these make sense at that points.

Everything works now, even with the custom selection parameter.


MajorMeerkatThe3rd:
@snabela related: https://github.com/shapediver/AppBuilderSdk/pull/94

snabela (approved):
Super @MajorMeerkatThe3rd, let's get this merged asap

PR #36, opened 2024-10-22
SS-8133 Set parameter values from query string

snabela:
https://shapediver.atlassian.net/browse/SS-8133

### DIFF
--- a/hooks/shapediver/appbuilder/useAppBuilderSettings.ts
+++ b/hooks/shapediver/appbuilder/useAppBuilderSettings.ts
@@ -10,6 +10,7 @@ import {
 	QUERYPARAM_DISABLEFALLBACKUI,
 	QUERYPARAM_MODELSTATEID,
 	QUERYPARAM_MODELVIEWURL,
+	QUERYPARAM_PARAMVALUE_PREFIX,
 	QUERYPARAM_PLATFORMURL,
 	QUERYPARAM_SETTINGSURL,
 	QUERYPARAM_SLUG,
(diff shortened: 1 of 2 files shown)


MajorMeerkatThe3rd (commented):
@snabela Are you sure about the usage of the "_" prefix? Not questioning the implementation, but just generally how it works.

Wouldn't a query parameter "parameters" make more sense where you can provide a JSON?

snabela:
> @snabela Are you sure about the usage of the "_" prefix? Not questioning the implementation, but just generally how it works.
> 
> Wouldn't a query parameter "parameters" make more sense where you can provide a JSON?

@MajorMeerkatThe3rd agreed that we could add a further option to specify a JSON object for multiple parameter values. However, so far the discussion with Edwin and Mathieu converged to the `_` prefix option. A JSON object isn't that simple to use via the query string for most users, because of the URL-encoding required.

MajorMeerkatThe3rd:
> > @snabela Are you sure about the usage of the "_" prefix? Not questioning the implementation, but just generally how it works.
> > Wouldn't a query parameter "parameters" make more sense where you can provide a JSON?
> 
> @MajorMeerkatThe3rd agreed that we could add a further option to specify a JSON object for multiple parameter values. However, so far the discussion with Edwin and Mathieu converged to the `_` prefix option. A JSON object isn't that simple to use via the query string for most users, because of the URL-encoding required.

@snabela Got it, makes sense then!

MajorMeerkatThe3rd (approved):

PR #39, opened 2024-11-04
SS-8091: generalization of useNodeInteractionData

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8091

@snabela I fixed the issue of the strict assignment of the loop length with this PR. Please have a look first into the useNodeInteractionData file where you can now see a separate way to call the hook. With this, we can create a single component per call and can even return the data of the hook with the form of a setState callback. Only important thing, these components have to be added to the dom (see ParameterSelectionComponent changes).

There are also some other changes in this commit to handle multiple sessions in a better way. Potentially there could have been issues if there are outputs with the same names in other sessions (as we never had the use case of something like that, it's obviously not an issue).

please don't merge it yet as it depends on the latest viewer version which is still a RC

### DIFF
--- a/components/shapediver/parameter/ParameterSelectionComponent.tsx
+++ b/components/shapediver/parameter/ParameterSelectionComponent.tsx
@@ -55,7 +55,7 @@ export default function ParameterSelectionComponent(props: PropsParameter) {
 	// get the viewport ID
 	const { viewportId } = useViewportId();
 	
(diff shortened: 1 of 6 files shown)


snabela (approved):
Many thanks @MajorMeerkatThe3rd, this looks great! 
At some we still might want to implement a `useOutputs` hook or so, but it's not very important.

PR #41, opened 2024-11-06
SS-8151 refresh JWT if expired

snabela:
https://shapediver.atlassian.net/browse/SS-8151


MajorMeerkatThe3rd (approved):
@snabela This should work now, thanks!

PR #43, opened 2024-11-06
SS-8140: fixed transformation issue and made some simplifications

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8140

There was an issue in which nodes where selected.

As I didn't find it at first, I did some "accidental" cleanup haha


snabela (approved):
Thank you @MajorMeerkatThe3rd

PR #44, opened 2024-11-07
SS-8148: prompt options and styling

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8148

implementation of the prompt and styling fixes and merges
now the wrapping of the text works as well

FYI: you accidentally created a branch with the same task number at some point. just to avoid the confusion 😂


snabela (approved):

PR #45, opened 2024-11-07
SS-8131 App Builder turns white in case of a computation timeout

snabela:
@MajorMeerkatThe3rd contains a bit more stuff than originally planned ;-) 
I refactored notifications, and established a mechanism for error reporting in general. Will be useful for several things in the near future. I will merge this right away, because the main repo also requires a dependent change.


MajorMeerkatThe3rd:
@snabela Thank you, makes a lot of sense!

PR #47, opened 2024-11-11
SS-8144: clear button for drawing and selection

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8144

had to do some changes to how the layout works as there cannot be a button within a button. Looks the same as before though 😂


snabela (approved):

PR #51, opened 2024-12-10
SS-8291: added widget to context and simplified defaults

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8291

preparation for the attribute vis widget

As for the Attribute Visualization we need a widget that is directly depending on the viewer, we need to also add context options for the widgets.

I also changed the context a bit so that you only have to provide the components that you want to change compared to the default.


snabela (approved):
@MajorMeerkatThe3rd many thanks, got it, makes sense to me!

PR #57, opened 2025-01-20
SS-8365: viewport icons in top middle on mobile

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8365

Small change from the description in the task (I'm not sure if you meant it that way or not), I implemented the useResponsiveValueSelector in the OverlayWrapper. 

Works as expected!


snabela (approved):
Many thanks @MajorMeerkatThe3rd ! I will merge it right away.

PR #58, opened 2025-01-20
SS-8380: Add svg support

LemToUp:
https://shapediver.atlassian.net/browse/SS-8380

### DIFF
--- a/components/shapediver/appbuilder/AppBuilderImage.tsx
+++ b/components/shapediver/appbuilder/AppBuilderImage.tsx
@@ -3,10 +3,15 @@ import { Image, ImageProps, MantineThemeComponent, Anchor, useProps } from "@man
 import { IAppBuilderWidgetPropsAnchor } from "../../../types/shapediver/appbuilder";
 import { AppBuilderContainerContext } from "../../../context/AppBuilderContext";
 import classes from "./AppBuilderImage.module.css";
+import Svg from "../../ui/Svg";
 
 type Props = IAppBuilderWidgetPropsAnchor;
 
-type ImageStyleProps = Pick<ImageProps, "radius" | "mah" | "maw"> & { fit?: "contain" | "scale-down", withBorder?: boolean};
+type ImageStyleProps = Pick<ImageProps, "radius" | "mah" | "maw"> & {
+	fit?: "contain" | "scale-down",
+	withBorder?: boolean,
+	isSvg?: boolean
+};
 
 type ImageNonStyleProps = Pick<ImageProps, "src"> & { alt?: string};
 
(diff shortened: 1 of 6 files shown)


components/ui/Svg.tsx
```diff
@@ -0,0 +1,47 @@
...
+			element.setAttribute("width", w !== undefined ? String(w) : "auto");
+			element.setAttribute("height", h !== undefined ? String(h) : "auto");
+			if (fit !== undefined) element.style["objectFit"] = fit;
+
+			if (maw !== undefined) element.style["maxWidth"] = String(maw);
+			if (mah !== undefined) element.style["maxHeight"] = String(mah);
+			if (className !== undefined) element.classList.add(className);
+
+			const radiusValue = (radius && Object.prototype.hasOwnProperty.call(radiusSizes, radius)) ? `var(--mantine-radius-${radius}, 0)` : radius;
+			if (radiusValue !== undefined) element.style["borderRadius"] = String(radiusValue);
```
snabela, line 38: @LemToUp please add some explanation about this to the code.

components/shapediver/appbuilder/AppBuilderImage.tsx
```diff
@@ -36,33 +41,43 @@ export function AppBuilderImageThemeProps(props: AppBuilderImageThemePropsType):
...
-		{element}
-	</Anchor> : element;
-
-	return elementWithAnchor;
+	const element = isSvg ? (
+		<Svg
+			{...contentProps}
+			srcUrl={rest.src}
+			mah={mah}
+			maw={maw}
```
snabela, line 64: @LemToUp did you not use the same logic for `mah` and `maw` on purpose? 
I mean this one: 

```
mah={!contain && orientation === "horizontal" ? mah ?? "100%" : undefined}
maw={!contain && orientation === "vertical" ? maw ?? "100%" : undefined}
```

types/shapediver/appbuildertypecheck.ts
```diff
@@ -141,6 +141,15 @@ const IAppBuilderWidgetPropsImageSchema = z.object({
...
 	target: z.string().default("_blank"),
+	isSvg: z.boolean().optional(),
+	maw: z.union([
+		z.number().optional(),
+		z.string().optional(),
+	]),
+	mah: z.union([
+		z.number().optional(),
+		z.string().optional(),
+	]),
```
snabela, line 152: @LemToUp let's drop maw and mah from here again

components/shapediver/appbuilder/AppBuilderImage.tsx
snabela: We must make the layout behavior exactly the same regardless of whether isSvg is true or false.

snabela: When removing the maw property from the json file, I get this: 

image

Probably that's why you added maw, but we must make it work without that.

snabela (changes requested):
Many thanks @LemToUp! Please review my comments, and let's talk about it.

components/ui/Svg.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+			setSvgElement(element);
+		});
+	}, [srcUrl]);
+
+	if (svgElement) return <div ref={ ref => ref?.appendChild(svgElement) }/>;
+
+	return <div/>;
```
snabela, line 46: @LemToUp maybe the div is preventing the logic for maw and mah to work as expected? Can we make it work without additional div?

types/shapediver/appbuildertypecheck.ts
LemToUp: Dropped

components/shapediver/appbuilder/widgets/AppBuilderImageWidgetComponent.tsx
```diff
@@ -2,36 +2,44 @@ import React from "react";
...
+	namespace: string,
+	isSvg?: boolean
 }
 
-export default function AppBuilderImageWidgetComponent(props: Props) {
+export default function AppBuilderImageWidgetComponent(props: Props & Pick<ImageProps, "maw" | "mah">) {
 	const {
 		alt,
+		maw,
+		mah,
```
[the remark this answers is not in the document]
LemToUp: Dropped

components/ui/Svg.tsx
LemToUp: It helps to support multiple inputs and svg content (like build in links) at the same time

PR #64, opened 2025-02-12
SS-8380: Add SVG unmount clean

LemToUp:
https://shapediver.atlassian.net/browse/SS-8380

### DIFF
--- a/components/ui/Svg.tsx
+++ b/components/ui/Svg.tsx
@@ -63,6 +63,8 @@ export default function Svg(props: SvgProps) {
 			});
 	}, [srcUrl]);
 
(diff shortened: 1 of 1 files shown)


components/ui/Svg.tsx
```diff
@@ -63,6 +63,8 @@ export default function Svg(props: SvgProps) {
 			});
 	}, [srcUrl]);
 
+	useEffect(() => () => setSvgElement(null), []);
+
```
snabela, line 67: @LemToUp could we combine this `useEffect` hook with the one above? That will also solve the case if `srcUrl` changes I guess.

snabela (approved):
Thank you @LemToUp !

PR #66, opened 2025-02-12
SS-8452: reset internal data once the parameter definition changed

MajorMeerkatThe3rd:
as discussed, FYI


snabela:

PR #74, opened 2025-02-24
SS-8494: Endless loop in App Builder 1.0.19

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8494

@snabela This was actually pretty straightforward. 

The useEffect hook was triggered whenever the parameter was updated, which caused another parameter update, etc etc

I now implemented a simple if statement if the value actually changed 😅


snabela (approved):
@MajorMeerkatThe3rd got it, many thanks!!!

PR #76, opened 2025-03-11
SS-8535: Refactor callback updates and looping hooks

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8535

@snabela I tried to split the commits up into logical chunks so that the reviewing process is easier for you:

- ddc352883e41dd40bbb00256553dc2b5c9dd4ba4: In this first commit I moved the updateCallbacks of sessions and outputs into the session store. You can see in the store also the subscribe function that keeps everything maintained. All just in one place, so I think that is now a very nice solution.
- c1fecfcf8a3e5adebaa77d303d997db7b21d6d71: In this second main commit I implemented new hooks to avoid the current looping mechanism. This works very nicely and I was able to clean up the code considerably (see the useRestrictions hook, that was a pain in the ass haha)
- 37cf529ca7356f2ead5cfe13cb037a80db50ebd4 and 20012c499fe12e4f3bd8baf38540fcaed077f8fe: smaller fixes with endless loops, classics
- 133ec5ee06cf7dc0b443784417e7dcad6d1b4e78: I restricted the support of all output usage to IDs, otherwise this can get confusing and can lead to errors, especially if there are multiple outputs with the same name.


MajorMeerkatThe3rd:
@snabela please review the last three commits
I implemented there what we talked about yesterday at the office and what you mentioned in your comments here

store/useShapeDiverStoreSession.ts
```diff
@@ -497,24 +547,87 @@ useShapeDiverStoreSession.subscribe((state, prevState) => {
...
+					outputsThatNeedUpdate.push({sessionId, outputId});
+				}
+			});
 		});
 	}
 
+	// remove duplicates from sessions that need update
+	sessionsThatNeedUpdate.filter(
+		(value, index, self) => self.indexOf(value) === index,
+	);
```
snabela, line 620: @MajorMeerkatThe3rd I think it should be like this: 
```
sessionsThatNeedUpdate = sessionsThatNeedUpdate.filter(
    (value, index, self) => self.indexOf(value) !== index,
);
```

store/useShapeDiverStoreSession.ts
```diff
@@ -497,24 +547,87 @@ useShapeDiverStoreSession.subscribe((state, prevState) => {
...
+	// remove duplicates from outputs that need update
+	outputsThatNeedUpdate.filter(
+		(value, index, self) =>
+			self.findIndex(
+				(item) =>
+					item.sessionId === value.sessionId &&
+					item.outputId === value.outputId,
+			) === index,
+	);
```
snabela, line 629: @MajorMeerkatThe3rd I think similar problem here
```
	// remove duplicates from outputs that need update
	outputsThatNeedUpdate = outputsThatNeedUpdate.filter(
		(value, index, self) =>
			self.findIndex(
				(item) =>
					item.sessionId === value.sessionId &&
					item.outputId === value.outputId,
			) !== index,
	);
```

store/useShapeDiverStoreSession.ts
MajorMeerkatThe3rd: @snabela I'm not so sure about that

image

snabela: @MajorMeerkatThe3rd sorry, had mixed it up in my mind. But we need to assign the result to sessionIds, no? filter returns a shallow copy, it doesn't modify the array in place.

MajorMeerkatThe3rd: @snabela Ah yes, you are right!

MajorMeerkatThe3rd: please see the latest commit

MajorMeerkatThe3rd:
@snabela All issues should be resolved now, if you are alright with it, I would merge to development tomorrow (as we just released a new version today).

snabela (approved):
Amazing, that's a biiiig step, many thanks @MajorMeerkatThe3rd !

PR #90, opened 2025-04-09
SS-8557 updates to the accept/reject behaviour

LemToUp:
https://shapediver.atlassian.net/browse/SS-8557

### DIFF
--- a/components/shapediver/parameter/ParameterWrapperComponent.module.css
+++ b/components/shapediver/parameter/ParameterWrapperComponent.module.css
@@ -1,3 +1,3 @@
 .wrapperModified {
-	background-color: var(--mantine-color-blue-light);
+	background-color: var(--mantine-primary-color-light) !important;
 }
--- a/components/shapediver/parameter/ParameterWrapperComponent.module.css
+++ b/components/shapediver/parameter/ParameterWrapperComponent.module.css
@@ -1,0 +1,3 @@
+.wrapperModified {
+	background-color: var(--mantine-color-blue-light);
+}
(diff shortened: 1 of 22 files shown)


components/shapediver/parameter/ParameterWrapperComponent.module.css
```diff
@@ -0,0 +1,3 @@
+.wrapperModified {
+	background-color: var(--mantine-color-blue-light);
```
snabela, line 2: @LemToUp let's use a variable that references a shade of the primary color, otherwise it will look strange if users use a different one

snabela: @LemToUp I tested this locally, and the color only works for me if I add `!important`.

components/shapediver/ui/ViewportAcceptRejectButtons.tsx
```diff
@@ -0,0 +1,136 @@
...
+				disabled={disableChangeControls}
+				{...buttonProps}
+			>
+				<Text {...textProps}>Reject</Text>
+			</Button>
+		</Group>
+	);
+}
+
+export default React.memo(ViewportAcceptRejectButtons);
```
snabela, line 136: @LemToUp ahhh, I didn't know about React.memo, we should probably start using it much more.

snabela (commented):
@LemToUp I have been testing locally, it works great! I am doing some refinements and will push the commits later today for your review.

snabela (approved):
@LemToUp many thanks, I will merge and deploy it now

PR #91, opened 2025-04-15
Fix async component loading

LemToUp:


snabela (approved):
@LemToUp ahh, many thanks!!!

PR #96, opened 2025-05-07
SS-8697: Add image and description to select image dropdown

LemToUp:
https://shapediver.atlassian.net/browse/SS-8697
@snabela that's prototype to check - do we need it or not.

### DIFF
--- a/components/shapediver/parameter/select/SelectImageDropDownComponent.module.css
+++ b/components/shapediver/parameter/select/SelectImageDropDownComponent.module.css
@@ -4,14 +4,16 @@
 }
 
 .input.inputSelected {
-  padding: 2px;
-  border-radius: 8px;
-  box-shadow: inset 0 0 0 2px var(--card-selected-color, var(--mantine-primary-color-filled));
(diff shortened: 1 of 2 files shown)


snabela (changes requested):
@LemToUp many thanks, this looks very good! 
A final comment regarding the border radius: I checked how it's defined for cards used by other types of dropdowns, and found this: 

image

Let's use the same CSS variable, which means we don't need to hardcode 8px radius anywhere.

LemToUp:
> @LemToUp many thanks, this looks very good! A final comment regarding the border radius: I checked how it's defined for cards used by other types of dropdowns, and found this:
> 
> image
> Let's use the same CSS variable, which means we don't need to hardcode 8px radius anywhere.

Good point! Done: https://github.com/shapediver/AppBuilderShared/pull/96/commits/1ccfee96aac8a6eb7dadee3907ebb69cfc170323

snabela (approved):
@LemToUp many thanks! I will merge this right away

PR #97, opened 2025-05-07
SS-8682: Markdown normalize

LemToUp:
https://shapediver.atlassian.net/browse/SS-8682


snabela (approved):
@LemToUp ah, great idea, many thanks!

PR #104, opened 2025-05-15
SS-8727: fixed issue with subscribe function

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8727

not an issue with zustand but with object nesting
as the nested object was not re-created, the prevState was already updated in the subscribe function
This only happened here because the state updates are called in a loop without breaking the javascript event loop


snabela:
> https://shapediver.atlassian.net/browse/SS-8727
> 
> not an issue with zustand but with object nesting as the nested object was not re-created, the prevState was already updated in the subscribe function This only happened here because the state updates are called in a loop without breaking the javascript event loop

@MajorMeerkatThe3rd got it, many thanks! 
Maybe we should start using `produce` more often (we already use it in some places)

PR #109, opened 2025-05-21
SS-7960: Add initial rendering transition

LemToUp:
https://shapediver.atlassian.net/browse/SS-7960
@snabela that is the second pr for the task - optional.


snabela (approved):
ahhh, thank you @LemToUp !

PR #114, opened 2025-06-03
SS-8748: use current jwtToken of the session

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8748

The issue was only in App Builder as there is a separate fetch function that always used the original token. So after the token was re-newed, the old token was still used for fetching.

I'm creatin a PR here, because I'm not sure why we use this fetching function in the first place. Is there a reason for it?
fetchFileWithToken can also receive a token to use


snabela (approved):
@MajorMeerkatThe3rd many thanks for fixing this! I think we didn't yet have `fetchFileWithToken` when we implemented this. Anyway, I would keep it as is for now.

PR #116, opened 2025-06-15
SS-8762 dynamic StringList parameters with images

snabela:
https://shapediver.atlassian.net/browse/SS-8762

### DIFF
--- a/types/shapediver/appbuildertypecheck.ts
+++ b/types/shapediver/appbuildertypecheck.ts
@@ -1,12 +1,42 @@
 import {IconTypeEnum} from "@AppBuilderShared/types/shapediver/icons";
 import {ShapeDiverResponseParameterStructure} from "@shapediver/api.geometry-api-dto-v2";
 import {
-	IDrawingParameterJsonSchema,
-	IInteractionParameterJsonSchema,
 	PARAMETER_TYPE,
 	PARAMETER_VISUALIZATION,
 } from "@shapediver/viewer.session";
 import {z} from "zod";
+
+// Zod type definition for SelectComponentType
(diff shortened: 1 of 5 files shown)


types/shapediver/appbuildertypecheck.ts
```diff
@@ -38,9 +68,7 @@ const IAppBuilderParameterDefinitionSchema = z.object({
 	tooltip: z.string().optional(),
 	displayname: z.string().optional(),
 	hidden: z.boolean(),
-	settings: z
-		.union([IInteractionParameterJsonSchema, IDrawingParameterJsonSchema])
-		.optional(),
+	settings: z.record(z.any()).optional(),
```
snabela, line 71: @MajorMeerkatThe3rd I removed this typecheck after trying to extend it by the new type. For some reason `zod` couldn't deal with this properly and returned an empty object. I plugged the type checking for the `ParameterSelectComponent` directly in there, we might do the same for the selection, drawing, and gumball parameters.

MajorMeerkatThe3rd: @snabela very interesting, as this definitely worked for the interaction parameters. But not an issue for me to move it there, I created a task: https://shapediver.atlassian.net/browse/SS-8783

MajorMeerkatThe3rd (approved):
@snabela looks good, thank you!

types/shapediver/appbuildertypecheck.ts
snabela: @MajorMeerkatThe3rd it seems to happen due to (partially) overlapping property names. Could be a bug in zod as well.

PR #118, opened 2025-06-24
SS-8730: Parameter history widget

LemToUp:
https://shapediver.atlassian.net/browse/SS-8730

### DIFF
--- a/hooks/shapediver/parameters/useParameterImportExport.ts
+++ b/hooks/shapediver/parameters/useParameterImportExport.ts
@@ -9,7 +9,9 @@ import {
 import {
 	filterAndValidateParameters,
 	generateParameterFeedback,
+	isImportParameterArray,
 } from "@AppBuilderShared/utils/parameters/parametersFilter";
+import {getParameterStates} from "@AppBuilderShared/utils/parameters/parameterStates";
 import {useCallback, useContext} from "react";
 import {useShallow} from "zustand/react/shallow";
 
@@ -36,20 +38,11 @@ export function useParameterImportExport(namespace: string) {
 	 * Export parameters as JSON file
 	 */
 	const exportParameters = useCallback(async () => {
-		// Get current parameter values from store
-		const parameters = useShapeDiverStoreParameters
-			.getState()
-			.getParameters(namespace);
-		const parameterArray: {id: string; value: any; name: string}[] = [];
-
-		Object.values(parameters).forEach((paramStore) => {
-			const param = paramStore.getState();
-			parameterArray.push({
-				id: param.definition.id,
-				value: param.state.execValue,
-				name: param.definition.name,
-			});
-		});
+		const parameterArray = getParameterStates(namespace).map((param) => ({
+			id: param.definition.id,
+			value: param.state.execValue,
+			name: param.definition.name,
+		}));
 
 		// Create JSON blob and download
 		const jsonContent = JSON.stringify({
@@ -134,12 +127,18 @@ export function useParameterImportExport(namespace: string) {
 					return;
 				}
 
-				const sessionParameters = useShapeDiverStoreParameters
-					.getState()
-					.getParameters(namespace);
+				if (!isImportParameterArray(importData.parameters)) {
+					const errorMessage =
+						"The schema of the parameters is not valid";
+					notifications.error({
+						message: errorMessage,
+					});
+					reject(new Error(errorMessage));
+					return;
+				}
 
 				const validationResult = filterAndValidateParameters(
-					sessionParameters,
+					getParameterStates(namespace),
 					importData.parameters,
 				);
 
@@ -179,15 +178,13 @@ export function useParameterImportExport(namespace: string) {
 	 * Reset parameters to default values
 	 */
 	const resetParameters = useCallback(async () => {
-		const parameters = useShapeDiverStoreParameters
-			.getState()
-			.getParameters(namespace);
-		const defaultValues: {[key: string]: any} = {};
-
-		Object.values(parameters).forEach((paramStore) => {
-			const param = paramStore.getState();
-			defaultValues[param.definition.id] = param.definition.defval;
-		});
+		const defaultValues = getParameterStates(namespace).reduce(
+			(acc, param) => {
+				acc[param.definition.id] = param.definition.defval;
+				return acc;
+			},
+			{} as Record<string, any>,
+		);
 
 		await batchParameterValueUpdate(namespace, defaultValues);
 
--- a/hooks/shapediver/useImportModelState.ts
+++ b/hooks/shapediver/useImportModelState.ts
@@ -8,6 +8,7 @@ import {
 	filterAndValidateModelStateParameters,
 	generateParameterFeedback,
 } from "@AppBuilderShared/utils/parameters/parametersFilter";
+import {getParameterStates} from "@AppBuilderShared/utils/parameters/parameterStates";
 import {useCallback, useContext, useState} from "react";
 import {useShallow} from "zustand/react/shallow";
 
@@ -75,13 +76,8 @@ export function useImportModelState(namespace: string) {
 				return false;
 			}
 
-			// Get session parameters and pass to utility function
-			const sessionParameters = useShapeDiverStoreParameters
-				.getState()
-				.getParameters(namespace);
-
 			const validationResult = filterAndValidateModelStateParameters(
-				sessionParameters,
+				getParameterStates(namespace),
 				parameters,
 			);
 
--- a/hooks/shapediver/stargate/useStargateExport.ts
+++ b/hooks/shapediver/stargate/useStargateExport.ts
@@ -4,11 +4,11 @@ import {useShapeDiverStoreStargate} from "@AppBuilderShared/store/useShapeDiverS
 import {IShapeDiverExportDefinition} from "@AppBuilderShared/types/shapediver/export";
 import {NetworkStatus} from "@AppBuilderShared/types/shapediver/stargate";
 import {exceptionWrapperAsync} from "@AppBuilderShared/utils/exceptionWrapper";
+import {getParameterStates} from "@AppBuilderShared/utils/parameters/parameterStates";
 import {ISdStargateExportFileResultEnum} from "@shapediver/sdk.stargate-sdk-v1";
 import {ShapeDiverResponseExportContent} from "@shapediver/viewer.session";
 import {useCallback, useContext, useEffect, useState} from "react";
 import {useShallow} from "zustand/react/shallow";
-import {useAllParameters} from "../parameters/useAllParameters";
 import {useStargateExportFile} from "./useStargateExportFile";
 import {ERROR_TYPE_INTERRUPTED} from "./useStargateGetData";
 
@@ -59,9 +59,6 @@ export const useStargateExport = ({
 	const [status, setStatus] = useState<ExportStatusEnum>(
 		ExportStatusEnum.notActive,
 	);
-
-	/** Get parameter stores of session */
-	const {parameters: parameterStores} = useAllParameters(sessionId);
 
 	const {exportFile} = useStargateExportFile();
 	const notifications = useContext(NotificationContext);
@@ -132,9 +129,8 @@ export const useStargateExport = ({
 			throw new Error("Current model not available");
 		}
 
-		const parameters = Object.values(parameterStores).reduce(
-			(acc, p_) => {
-				const p = p_.getState();
+		const parameters = getParameterStates(sessionId).reduce(
+			(acc, p) => {
 				acc[p.definition.id] = p.state.stringExecValue();
 				return acc;
 			},
@@ -203,7 +199,7 @@ export const useStargateExport = ({
 					],
 			});
 		}
-	}, [parameterStores, exportFile, exportId, contentIndex]);
+	}, [sessionId, exportFile, exportId, contentIndex]);
 
 	return {
 		isWaiting,
--- a/hooks/shapediver/stargate/useStargateOutput.ts
+++ b/hooks/shapediver/stargate/useStargateOutput.ts
@@ -5,6 +5,7 @@ import {useShapeDiverStoreStargate} from "@AppBuilderShared/store/useShapeDiverS
 import {IShapeDiverOutputDefinition} from "@AppBuilderShared/types/shapediver/output";
 import {NetworkStatus} from "@AppBuilderShared/types/shapediver/stargate";
 import {exceptionWrapperAsync} from "@AppBuilderShared/utils/exceptionWrapper";
+import {getParameterStates} from "@AppBuilderShared/utils/parameters/parameterStates";
 import {ISdStargateBakeDataResultEnum} from "@shapediver/sdk.stargate-sdk-v1";
 import {
 	ITreeNode,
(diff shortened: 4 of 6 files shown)


snabela:
@LemToUp before starting with this refactoring, did you try to call `history.forward()` and `history.backward()`? This should work. 
https://developer.mozilla.org/en-US/docs/Web/API/History_API#moving_forward_and_backward

snabela:
> @LemToUp before starting with this refactoring, did you try to call `history.forward()` and `history.backward()`? This should work. https://developer.mozilla.org/en-US/docs/Web/API/History_API#moving_forward_and_backward

I tried to do this in the browser console, works as expected. Please also see the task spec where I mentioned this: 
image

hooks/shapediver/parameters/useParameterHistory.ts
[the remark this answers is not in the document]
LemToUp: @snabela it uses viewer's built in functional now.

types/store/shapediverStoreParameters.ts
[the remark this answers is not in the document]
LemToUp: Rolled back.

store/useShapeDiverStoreParameters.ts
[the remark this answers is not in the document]
LemToUp: Rolled back.

hooks/shapediver/useModelState.ts
snabela, line 1: @LemToUp I suggest to split this hook into two: 

1. Instead of the new hook `useModelState`, we can use the existing hook called `useCreateModelState`
2. Move `importModelState` to a new file `useImportModelState`

components/ui/Icon.tsx
snabela, line 1: @LemToUp I have adapted some of the changes you did in the other PR, to keep the naming consistent. Please do the same here: https://github.com/shapediver/AppBuilderShared/pull/115/files#diff-ad4ef36c6c74842945e46dbdabd341567f4a580ed537687bf58c715b39c3e545

components/shapediver/viewport/ViewportHistoryButtons.tsx
```diff
@@ -0,0 +1,260 @@
...
+	variantDisabled?: string;
+	/**
+	 * Style for individual icons
+	 */
+	iconStyle?: MantineStyleProp;
+	/**
+	 * Optional list of session IDs to check for pending parameter changes.
+	 * If provided, buttons will be disabled when there are pending changes for these sessions.
+	 */
+	sessionIds?: string[];
```
snabela, line 46: @mathieu-shapediver do you think we will need this (disable history buttons if there are pending parameter changes)? I would avoid it, because the browser history buttons have the exact same functionality and can't be disabled as well.

snabela (commented):
@LemToUp many thanks! I left some comments.

hooks/shapediver/parameters/useParameterImportExport.ts
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Hook for managing parameter import/export and reset functionality.
+ */
+export function useParameterImportExport(namespace: string) {
+	const {sessions} = useShapeDiverStoreSession(
+		useShallow((state) => ({
+			sessions: state.sessions,
+		})),
+	);
```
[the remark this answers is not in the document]
LemToUp: Removed

components/shapediver/viewport/ViewportHistoryButtons.tsx
```diff
@@ -0,0 +1,272 @@
...
+	const parameterChanges = useShapeDiverStoreParameters(
+		useCallback(
+			(state) =>
+				Object.keys(state.parameterChanges).reduce((acc, id) => {
+					acc.push(state.parameterChanges[id]);
+					return acc;
+				}, [] as IParameterChanges[]),
+			[],
+		),
+	);
```
snabela, line 129: @LemToUp small optimization possibility: `executing` can probably be defined based on `parameterChanges`. 
Great idea to use `useCallback` for the state selector. I wasn't aware that this is possible.

components/ui/Hint.tsx
```diff
@@ -0,0 +1,97 @@
...
+		useProps("Hint", defaultStyleProps, props);
+
+	return (
+		<Group {...containerGroupProps}>
+			<Group {...groupProps}>
+				<Icon {...iconProps} />
+				<Text {...textProps}>{title}</Text>
+			</Group>
+			<Button {...buttonProps} href={docLink}>
+				READ MORE
```
snabela, line 93: @mathieu-shapediver same here

hooks/shapediver/viewer/useViewportHistory.ts
snabela, line 1: @LemToUp I researched under what circumstances `history.back()` or forward would throw. It is extremely unlikely to happen, therefore let's not catch it. This also means we can drop the `isLoading` state, it's not needed.

hooks/shapediver/useImportModelState.ts
```diff
@@ -0,0 +1,101 @@
...
+				return true;
+			} catch (error) {
+				console.error("Error importing model state:", error);
+				notifications.error({
+					message: "Failed to import model state",
+				});
+				return false;
+			} finally {
+				setIsLoading(false);
+			}
```
snabela, line 80: @LemToUp a few comments: 

  * If we catch errors here, we need to report them to sentry as well. Therefore, instead of `console.error`, let's use the `ErrorReportingContext`
  *  In case of success let's use `Model state {trimmedId} imported successfully`
  * We should not manipulate parameter values using the session API directly. We must only use the parameter stores. You can use the geometry backend SDK to fetch the data of the model state directly, and then apply the values to the parameter store. https://sddev2.eu-central-1.shapediver.com/api/v2/docs/#/model-state

components/shapediver/viewport/ViewportHistoryButtons.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+const defaultProps: Required<Omit<Props, "namespace">> = {
+	style: {display: "flex", gap: "4px"},
+	size: 32,
+	color: "black",
+	colorDisabled: "grey",
+	variant: "subtle",
+	variantDisabled: "transparent",
+	iconStyle: {m: "3px"},
+};
```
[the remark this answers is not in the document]
MajorMeerkatThe3rd: @snabela I agree with using rem, If you want to I can move through the currently usages of `px` and see where it makes sense to replace it when I do this task: https://shapediver.atlassian.net/browse/SS-8784

hooks/shapediver/parameters/useParameterImportExport.ts
```diff
@@ -0,0 +1,162 @@
...
+			// Get current parameter values from store
+			const parameters = useShapeDiverStoreParameters
+				.getState()
+				.getParameters(namespace);
+			const parameterValues: {[key: string]: any} = {};
+
+			Object.values(parameters).forEach((paramStore) => {
+				const param = paramStore.getState();
+				parameterValues[param.definition.id] = param.state.uiValue;
+			});
```
snabela, line 32: @LemToUp @mathieu-shapediver a question about the "export parameters as JSON" button: We disable this button if there are pending parameter changes, correct? We should not use `param.state.uiValue` here, because it's the value shown in the UI, which can be different from the "executed value"  `execValue`.

hooks/shapediver/parameters/useParameterImportExport.ts
```diff
@@ -0,0 +1,162 @@
...
+			const link = document.createElement("a");
+			link.href = url;
+			link.download = `parameters_${namespace}_${new Date().toISOString().split("T")[0]}.json`;
+			document.body.appendChild(link);
+			link.click();
+			document.body.removeChild(link);
+			URL.revokeObjectURL(url);
+
+			notifications.success({
+				message: "Parameters exported successfully",
```
snabela, line 49: Let's use _Parameter values exported successfully`

hooks/shapediver/parameters/useParameterImportExport.ts
```diff
@@ -0,0 +1,162 @@
...
+			});
+		} catch (error) {
+			console.error("Error exporting parameters:", error);
+			notifications.error({
+				message: "Failed to export parameters",
+			});
+		}
```
snabela, line 56: @LemToUp instead of console.error, let's report to Sentry

hooks/shapediver/parameters/useParameterImportExport.ts
```diff
@@ -0,0 +1,162 @@
...
+					resolve();
+				} catch (error) {
+					console.error("Error importing parameters:", error);
+					notifications.error({
+						message:
+							"Failed to import parameters: " +
+							(error as Error).message,
+					});
+					reject(error);
+				}
```
snabela, line 116: @LemToUp related to this [comment](https://github.com/shapediver/AppBuilderShared/pull/118/files#r2171041826):

  * Parameters should be searched first by id, then by name
  * In case some of the parameters could not be found, we should show a warning notification listing all the names of the parameters that could not be found. 
  * In case all of the parameters could not be found, show an error notification. 
  * Let's not use try-catch here, because we want errors to be reported to sentry

hooks/shapediver/viewer/useViewportHistory.ts
LemToUp: https://github.com/shapediver/AppBuilderShared/commit/633914a7c058a1d1e6b9351bca125311e271129b

components/shapediver/viewport/ViewportHistoryButtons.tsx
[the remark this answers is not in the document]
LemToUp: @snabela I've leaved the border in px.
That will be confusing if the border will be font size related.
https://github.com/shapediver/AppBuilderShared/commit/633914a7c058a1d1e6b9351bca125311e271129b

components/shapediver/ui/ImportModelStateDialog.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+	return (
+		<ModalBase
+			opened={opened}
+			onClose={handleClose}
+			onConfirm={handleImport}
+			title="Load a model state"
+			isLoading={isLoading}
+			cancelBtnTitle="CANCEL"
+			confirmBtnTitle="LOAD"
```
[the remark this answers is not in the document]
mathieu-shapediver: @snabela Agreed!

components/ui/Hint.tsx
snabela: @LemToUp let's change it to _Read more_

hooks/shapediver/parameters/useParameterImportExport.ts
LemToUp: 1c02dbe28bf8225b435d8a59520c4883170edfe8

hooks/shapediver/parameters/useParameterImportExport.ts
LemToUp: @snabela Do you remember, is the `model.id` accessible at the app ?

components/shapediver/ui/ImportModelStateDialog.tsx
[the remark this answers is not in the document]
LemToUp: fb56704c10e700228588c74f18012bf54be347bc

components/ui/Hint.tsx
LemToUp: fb56704c10e700228588c74f18012bf54be347bc

hooks/shapediver/parameters/useParameterImportExport.ts
snabela: @LemToUp yes you can use `useShapeDiverStorePlatform.currentModel`. However, it might be undefined (e.g. if no slug was used). In that case we simply don't include the model id in the JSON.

components/shapediver/viewport/ViewportHistoryButtons.tsx
snabela: @mathieu-shapediver ok sounds good, let's keep it! 
Note: The browser history buttons can't be blocked, let's keep the current behavior.

hooks/shapediver/parameters/useParameterImportExport.ts
LemToUp: Done: b7f7b4a459c953728a609dd71d8e03f8f56a2edf

hooks/shapediver/useImportModelState.ts
LemToUp: Done.

hooks/shapediver/parameters/useParameterImportExport.ts
LemToUp: Done: 59b8e1d93fb87c794eaa1a9ce41ba25379cd71e0

hooks/shapediver/useImportModelState.ts
```diff
@@ -51,19 +50,21 @@ export function useImportModelState(namespace: string) {
 
 			setIsLoading(true);
 			try {
-				await sessionApi.customizeWithModelState(trimmedId);
+				// TODO: get parameters from state id
```
MajorMeerkatThe3rd, line 53: @LemToUp @snabela I just deployed viewer version `3.11.3` which exposes the new function `getModelState` on the SessionApi

LemToUp: Done: 992ab5c195f4d8c161db216444ee7a75e7cbac7d

hooks/shapediver/useImportModelState.ts
```diff
@@ -0,0 +1,92 @@
...
+	 * Import a model state by ID
+	 */
+	const importModelState = useCallback(
+		async (modelStateId: string) => {
+			if (!sessionApi) {
+				notifications.error({
+					message: "Session not available",
+				});
+				return false;
+			}
```
snabela, line 41: @LemToUp if this happens it's a developer error, i.e. we don't need this

components/shapediver/modelState/ModelStateNotificationCreated.tsx
```diff
@@ -0,0 +1,42 @@
...
+		navigator.clipboard.writeText(link);
+		notifications.success({
+			message: "State link copied to clipboard!",
+		});
+	}, [modelStateId, link, notifications]);
+
+	return (
+		<>
+			Model state with ID {modelStateId} has been saved.
+			<Anchor ml={4} size="sm" component="span" onClick={copyStateLink}>
```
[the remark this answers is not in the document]
snabela: @LemToUp let's revert these commits, and start a separate PR, but only once this one has been finished and merged. It's getting too much.

hooks/shapediver/useImportModelState.ts
```diff
@@ -49,38 +51,63 @@ export function useImportModelState(namespace: string) {
...
+				notifications[feedback.type]({
+					message: feedback.message,
+				});
+				return false;
+			}
+
+			await batchParameterValueUpdate(
+				namespace,
+				validationResult.validParameters,
+				!validationResult.acceptRejectMode,
```
snabela, line 97: @LemToUp quick question: what is the reason we are skipping creating a history entry if all parameters are not using `acceptRejectMode`?

snabela: bug found

PR #119, opened 2025-06-25
SS-8789: Extend markdown by sup, sub and ins

LemToUp:
https://shapediver.atlassian.net/browse/SS-8789


snabela (approved):
Ah, great, many thanks @LemToUp !

PR #127, opened 2025-07-01
SS-8784: Improve theme capabilities for currently unsupported properties

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8784

@snabela I just now only did this for the AV, please confirm if you like it that way before I also do that for other files

### DIFF
--- a/components/shapediver/appbuilder/widgets/AppBuilderAreaChartWidgetComponent.tsx
+++ b/components/shapediver/appbuilder/widgets/AppBuilderAreaChartWidgetComponent.tsx
@@ -2,25 +2,60 @@ import {
 	convertChartData,
 	IAppBuilderWidgetPropsAreaChart,
 } from "@AppBuilderShared/types/shapediver/appbuildercharts";
-import {AreaChart} from "@mantine/charts";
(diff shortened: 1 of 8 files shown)


snabela (approved):
Thank you @MajorMeerkatThe3rd !
Small doubt: Did we drop the `order` property of the `Title` components on purpose, correct?

snabela:
> > Thank you @MajorMeerkatThe3rd ! Small doubt: Did we drop the `order` property of the `Title` components on purpose, correct?
> 
> @snabela As it was the same order for all components where the Title was used, I set is as a general default: [9e0503b#diff-e18abaf37f3780a6decb3ec191b262583dcced536d89969b083d483c2a2f7c8bR359](https://github.com/shapediver/AppBuilderShared/commit/9e0503b2dd9c66d56691b5555f9ac6e78a5e8dd4#diff-e18abaf37f3780a6decb3ec191b262583dcced536d89969b083d483c2a2f7c8bR359)
> 
> Let me know if you want that on a per-component basis

Got it, makes sense!

snabela (approved):

PR #133, opened 2025-07-04
SS-8696: 3D anchor containers (incl justification & allowPointerEvents)

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8696

@snabela An initial draft PR so that you can already have a look and request changes. Device testing still pending.

Some notes:

- I managed to keep this super simple by directly using the AppBuilderContainerComponent (see AppBuilderPage logic)
- For the anchor justification I'm using the same justification as is used in GH for the TextTag3D
- The initially weird looking group structure in the ViewportAnchor is needed to ensure that the overflow is managed correctly. This took some time to figure out but now it works really well.
- Some styling improvements will still be needed in the future for other combinations of tabs and widgets
- For now I set the maxWidth: "var(--app-shell-navbar-width)", how do you feel about that?

### DIFF
--- a/types/shapediver/appbuilder.ts
+++ b/types/shapediver/appbuilder.ts
@@ -10,6 +10,7 @@ import {IShapeDiverParameterDefinition} from "@AppBuilderShared/types/shapediver
 import {SessionCreateDto} from "@AppBuilderShared/types/store/shapediverStoreSession";
 import {MantineColor} from "@mantine/core";
 import {Gradient} from "@shapediver/viewer.features.attribute-visualization";
+import {TAG3D_JUSTIFICATION} from "@shapediver/viewer.session";
 
 /** Type used for parameter definitions */
 export type IAppBuilderParameterDefinition = IShapeDiverParameterDefinition & {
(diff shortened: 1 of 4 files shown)


types/shapediver/appbuilder.ts
```diff
@@ -414,14 +415,33 @@ export interface IAppBuilderTab {
...
+	previewIcon?: IconType;
+};
 
 /**
  * A container for UI elements
  */
 export interface IAppBuilderContainer {
 	/** Name of the container. */
-	name: AppBuilderContainerNameType;
+	name: AppBuilderContainerNameType | AppBuilderContainerLocationType;
```
snabela, line 444: @MajorMeerkatThe3rd sorry for the late review! A comment about this combined type: It's a pain to achieve such types in .NET, because serializers need to be implemented. Therefore I suggest to switch to a schema that uses a further optional property for storing these properties. We could name it `props`. The `name` enum would need to be extended to allow arbitrary string values. What do you think?

MajorMeerkatThe3rd: @snabela We could actually use the `name` still as an enum. In that case we could have a new name `anchor3d` and corresponding typed `props`. Then on the typescript side we could use the `name` property for a `discriminatedUnion` check regarding the `props`.
I think that might be a nice solution which is then clearly structured. Any objections?

MajorMeerkatThe3rd:
@snabela This is ready for review now 🎉 
We can discuss later at the office, but I think I managed to clean everything up really nicely.

types/store/shapediverStoreViewportAnchors.ts
```diff
@@ -0,0 +1,47 @@
+export interface IAnchor {
+	distance: number;
+	showContent: boolean;
+	setShowContent: (showContent: boolean) => void;
+	setZIndex: (zIndex: number) => void;
+	id: string;
+}
```
snabela, line 7: @MajorMeerkatThe3rd let's add documentation here

snabela (dismissed):
As discussed let's try fullscreen mode

types/store/shapediverStoreViewportAnchors.ts
MajorMeerkatThe3rd: Done!

PR #135, opened 2025-07-08
SS-8818 optional "step" setting for number inputs

snabela:
https://shapediver.atlassian.net/browse/SS-8818


snabela:
> @snabela Looks great. Should generally some of this move into the viewer eventually?

For now we are just allowing this for dynamically defined parameters. Likely this will stay so. If we ever implement it for our common parameters, we will have to update the geometry backend DTOs and SDK as a first step.

PR #140, opened 2025-07-15
SS-8827: Viewport buttons refactoring

LemToUp:
https://shapediver.atlassian.net/browse/SS-8827


components/shapediver/viewport/buttons/types.ts
```diff
@@ -0,0 +1,16 @@
...
+export interface CommonButtonProps {
+	size?: number;
+	color?: string;
+	colorDisabled?: string;
+	variant?: string;
+	variantDisabled?: string;
+	iconStyle?: MantineStyleProp;
+}
+
+export const IconSize = 32;
```
MajorMeerkatThe3rd, line 12: @snabela @LemToUp This should probably be adapted to make it themeable, right?
Currently this is not like we did it for other component, so maybe have a look at that to keep the coding principle the same.

pages/appbuilder/AppBuilderPage.tsx
```diff
@@ -318,23 +317,15 @@ export default function AppBuilderPage(props: Partial<Props>) {
 					>
 						{ViewportOverlayWrapper && (
 							<>
-								<ViewportOverlayWrapper>
-									{ViewportIcons && <ViewportIcons />}
-								</ViewportOverlayWrapper>
+								{ViewportIcons && (
+									<ViewportIcons namespace={namespace} />
```
MajorMeerkatThe3rd, line 321: @LemToUp general question, why was previously the ViewportOverlayWrapper needed and isn't anymore?

MajorMeerkatThe3rd (commented):
@LemToUp I did an initial review and left some comments, Alex might also want to have a look at this after this vacation

pages/appbuilder/AppBuilderPage.tsx
LemToUp: @MajorMeerkatThe3rd one of the requirements - configurable buttons position.
So I replaced ViewportOverlayWrapper into the component for that.
At the same time - there is possible to configure buttons direction (horizontal / vertical)

components/shapediver/viewport/buttons/types.ts
LemToUp: Good point @MajorMeerkatThe3rd 
Default size removed.

LemToUp: @MajorMeerkatThe3rd it may be managed via custom theme properties.
Default colors may be used too.
But the default variant - looks bad.

pages/appbuilder/AppBuilderPage.tsx
LemToUp: @MajorMeerkatThe3rd based on the recent discussion:
https://github.com/shapediver/AppBuilderShared/pull/140/commits/bb8de3d2531eecb50b00468cc95af56c392e6c2e

components/shapediver/viewport/ViewportIcons.tsx
```diff
@@ -1,289 +1,183 @@
...
-				</Modal>
-			)}
-
-			{enableZoomBtn && (
-				<TooltipWrapper label="Zoom extents">
-					<ActionIcon
-						onClick={zoomClickHandler}
-						size={size}
+		<ViewportOverlayWrapper position={position} offset={offset}>
+			<Paper style={style} shadow={shadow} py={py} px={px}>
```
snabela, line 111: @LemToUp given that we are already refactoring the theme properties of this component (and potentially breaking existing themes), what about changing this to `{...viewportOverlayProps}` respectively `{...paperProps}`? This seems easier to maintain in the future and we don't need to refactor should we ever extend `ViewportOverlayWrapper` by further properties.

components/shapediver/viewport/buttons/types.ts
```diff
@@ -0,0 +1,16 @@
+import {MantineStyleProp} from "@mantine/core";
+
+export interface CommonButtonProps {
+	size?: number;
+	color?: string;
+	colorDisabled?: string;
```
snabela, line 6: It seems `color` and `colorDisabled` aren't used?

snabela (changes requested):
@LemToUp I left a few comments
@MajorMeerkatThe3rd FYI

components/shapediver/viewport/buttons/types.ts
```diff
@@ -0,0 +1,16 @@
...
+	variant?: string;
+	variantDisabled?: string;
+	iconStyle?: MantineStyleProp;
```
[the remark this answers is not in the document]
LemToUp: https://shapediver.atlassian.net/browse/SS-8827?focusedCommentId=43702

snabela (approved):
MAny thanks @LemToUp !

PR #145, opened 2025-07-22
SS-8545: Add anchor elements when an object is clicked in AV mode

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8545

FYI: base for this is the Anchors branch, as it builds on this.
I decided against making this a widget for now as I'm not sure if it would ever get used.

The separation is done by having a hook `useAttributeSelection`, which handles the selection of the attributes depending on certain criteria. This can then also be re-used when we implement a global functionality for clicking on objects and letting people inspect the attribute data (as discussed with Mathieu).

The `SelectedAttributeComponent` is a relatively simple component that puts everything in a table and just has some logic regarding which objects should be shown etc.


snabela (dismissed):
Got it, many thanks @MajorMeerkatThe3rd !

PR #147, opened 2025-07-22
SS-8860 export/output button interceptor

snabela:
https://shapediver.atlassian.net/browse/SS-8860


components/shapediver/outputs/OutputChunkLabelComponent.tsx
```diff
@@ -47,11 +49,16 @@ export default function OutputChunkLabelComponent(
...
+	return (
+		<Group justify="space-between" w="100%" wrap="nowrap">
+			{tooltip ? (
+				<TooltipWrapper label={tooltip} position="top">
+					{labelcomp}
+				</TooltipWrapper>
+			) : (
+				labelcomp
+			)}
+			{rightSection}
```
MajorMeerkatThe3rd, line 61: @snabela Will this rightSection now always be present here? In all use cases?

MajorMeerkatThe3rd (approved):
@snabela Thank you, only left a small comment, but approved already

components/shapediver/outputs/OutputChunkLabelComponent.tsx
snabela: we will use it for PERI only, there will be no visible change for the main distribution

snabela:
@LemToUp @MajorMeerkatThe3rd merging this

PR #148, opened 2025-07-22
SS-8854: fixed accidental showing of accept/reject

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8854

the issue was that one of the interaction parameters updated a parameter value to the same value that was already set. I handled this in the useParameterComponentCommons, let me know if you want me to move that to the separate interaction parameters instead. I think it makes sense here, either way.


MajorMeerkatThe3rd:
@snabela Please review, I now did this only for the interaction parameters where this can happen.

PR #153, opened 2025-07-25
SS-8818 console warning in case of invalid parameter value

snabela:
https://shapediver.atlassian.net/browse/SS-8818?focusedCommentId=43957


MajorMeerkatThe3rd (approved):
@snabela That makes sense. What do you think about also using a warning notification here? I think that might make sense, as otherwise I already see users asking why something is not working because they won't open the console.

PR #157, opened 2025-07-30
SS-8827: Mobile touch events and modal visibility

LemToUp:
https://shapediver.atlassian.net/browse/SS-8827?focusedCommentId=44179
@MajorMeerkatThe3rd I've fixed 1st and 3rd items. (Except On mobile (android and iOS) the buttons only get activated on a single press)


MajorMeerkatThe3rd (approved):
@LemToUp Thank you! This works now, I just removed the onTouch events from the theme. This cannot be set by the user anyway and can be hardcoded in this case!

@snabela FYI

PR #163, opened 2025-08-05
SS-8897: use iframe settings in app builder for saved states and JSON

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8897

@snabela This works now as expected. 
One thing for you to check: hooks/shapediver/appbuilder/useResolveAppBuilderSettings.ts
In this file I now use just the iframe settings. For each of those settings there are two versions, non-iframe and iframe. I this the correct one for App Builder usage?

See this comment for the deployed example on a feature branch:
https://shapediver.atlassian.net/browse/SS-8897?focusedCommentId=44681


snabela (approved):
Great, many thanks @MajorMeerkatThe3rd !

PR #164, opened 2025-08-05
SS-8899: Duplicate platform calls

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8899

@snabela took a bit to figure this out. The main issue was that the sessions were not properly memoized, especially when a JSON was loaded. I now wait for the JSON to load and then continue.

I also did some further improvements. As this is a critical part, I think a review makes sense


snabela (approved):
Got it, many thanks @MajorMeerkatThe3rd !!!

PR #165, opened 2025-08-06
SS-8898 parameter batch update: warn instead of throw

snabela:
https://shapediver.atlassian.net/browse/SS-8898

The last PR before we can deploy App Builder 1.3


snabela:
> @snabela This makes sense! I think this is again a topic where we could show real notifications (not just in the console). Should I add a task to unify notifications/warnings/error? We can then talk more about this in a spec phase, not urgent

Agreed! We have a task for this in release 1.4: https://shapediver.atlassian.net/browse/SS-8825
I will add a link to this PR from the task.

PR #168, opened 2025-08-12
SS-8869: mobile fallback behavior for viewport anchors

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8869

@snabela I had to create a store as I needed to add the content of the viewport anchors to it.
In the end I only implemented Stage 1 & 2 as I didn't see a need for Stage 3 yet.


types/store/shapediverStoreStandardContainers.ts
```diff
@@ -0,0 +1,98 @@
...
+	/**
+	 * The default containers for each position.
+	 * These are the containers that are either specified by the user or created as a fallback.
+	 *
+	 * We store the definition of these containers here, to be able to adjust them later if needed.
+	 */
+	defaultContainers: Record<
+		AppBuilderStandardContainerNameType,
+		IAppBuilderContainer | undefined
+	>;
```
snabela, line 38: Suggestion for doc update: 

The default container state for each of the standard container names. 
This container state is defined by the `AppBuilder` output of the controller session, or the fallback logic in case no `AppBuilder` output exists.

types/store/shapediverStoreStandardContainers.ts
```diff
@@ -0,0 +1,98 @@
...
+	setDefaultContainer: (
+		position: AppBuilderStandardContainerNameType,
+		container: IAppBuilderContainer | undefined,
+	) => void;
+
+	/**
+	 * Reset the default container definitions to their initial state.
+	 * This will remove any user-defined containers and restore the default containers.
+	 */
+	resetDefaultContainers: () => void;
```
snabela, line 77: maybe rename this to `resetToDefaultContainers`

types/store/shapediverStoreStandardContainers.ts
```diff
@@ -0,0 +1,98 @@
...
+		container: JSX.Element,
+	) => string;
+
+	/**
+	 * Remove an additional container for a specific position.
+	 *
+	 * @param token - The unique token representing the added container.
+	 * @returns A boolean indicating whether the container was successfully removed.
+	 */
+	removeAdditionalContainer: (token: string) => boolean;
```
snabela, line 97: I suggest renaming this to `removeAdditionalContainerContent` (or `Element`)

hooks/shapediver/appbuilder/useAppBuilderStandardContainers.tsx
```diff
@@ -0,0 +1,182 @@
...
+		// Reset containers first
+		resetDefaultContainers();
+
+		if (appBuilderData?.containers) {
+			// Set containers from appBuilderData
+			appBuilderData.containers.forEach((container) => {
+				if (isStandardContainer(container)) {
+					setDefaultContainer(container.name, container);
+				}
+			});
```
snabela, line 106: Optimization idea: Likely we can implement a `setDefaultContainers` state function which does everything in a single `set` call in the store, thereby avoiding some re-rendering.

PR #169, opened 2025-08-12
SS-8909/SS-8903: platform settings implementation

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8909
https://shapediver.atlassian.net/browse/SS-8903

Two tasks in one as they were very related. 
I know use the settings as specified in the first task.

If only ticket and modelViewUrl is provided, I use the settings stored on the viewer.

Can be tested on branch `SS-8909` by using models from the dev platform where we now also store these settings in the viewer as a fallback.


snabela (approved):
This is great, many thanks @MajorMeerkatThe3rd !

PR #172, opened 2025-08-14
Task/ss 8865

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8865

ok another try, I now switched to a different library for the icons so that they can be loaded on demand 
Please review, everything looks exactly the same visually


snabela (approved):
Nice! Many thanks @MajorMeerkatThe3rd!

PR #173, opened 2025-08-14
SS-8935: Fix import functionality issue for NordicBim

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8935

@snabela Just a short review if that makes sense to you.

The issue was that within the json that was imported, there were custom parameters with different values. As the getCustomParameterValues function returned the current values, these values were never updated and overwritten.

### DIFF
--- a/hooks/shapediver/appbuilder/useAppBuilderCustomParameters.ts
+++ b/hooks/shapediver/appbuilder/useAppBuilderCustomParameters.ts
@@ -187,7 +187,17 @@ export function useAppBuilderCustomParameters(props: Props) {
 			setPreExecutionHook(namespace, async (_values) => {
 				const values = {..._values};
 				const customValues = getCustomParameterValues();
-				const json = JSON.stringify(customValues);
+
+				// if within the values that should be set, there is a AppBuilder parameter defined
+				// we don't use the custom values, but the values within that parameter
+				// as the custom values are stale in that case
+				let json;
+				if (appBuilderParam && values[appBuilderParam.definition.id]) {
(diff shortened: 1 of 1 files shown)


hooks/shapediver/appbuilder/useAppBuilderCustomParameters.ts
```diff
@@ -187,7 +187,17 @@ export function useAppBuilderCustomParameters(props: Props) {
...
+				// if within the values that should be set, there is a AppBuilder parameter defined
+				// we don't use the custom values, but the values within that parameter
+				// as the custom values are stale in that case
+				let json;
+				if (appBuilderParam && values[appBuilderParam.definition.id]) {
+					json = values[appBuilderParam.definition.id];
+				} else {
+					json = JSON.stringify(customValues);
+				}
+
```
snabela, line 200: @MajorMeerkatThe3rd I have doubts about this fix. Let's discuss it on Monday. Let's describe here the cases in which `values` can contain a value for the AppBuilder parameter. 
This needs to be tested very carefully, for example, concerning the parameter history functionality for apps that contain dynamic parameters.

MajorMeerkatThe3rd: @snabela Got it, I thought there might be more to it. To be discussed

MajorMeerkatThe3rd: Discussed with @snabela, can be merged like that

PR #174, opened 2025-08-18
SS-8888: Refactor viewport icons/buttons/menus

LemToUp:
https://shapediver.atlassian.net/browse/SS-8888

### DIFF
--- a/store/useShapeDiverViewportIconsStore.ts
+++ b/store/useShapeDiverViewportIconsStore.ts
@@ -229,6 +229,7 @@ export const useShapeDiverViewportIconsStore =
 						`icons:clear ${viewportId}`,
 					);
 				},
+				/** Manage common props and render the viewport icons, based on the viewport icons state (buttons and groups) */
 				render: (
 					viewportId: string,
 					buttonContext: ButtonRenderContext,


components/shapediver/viewport/buttons/ViewportIconButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+	disabled?: boolean;
+	onClick?: () => void;
+}
+
+export default function ViewportIconButton({
+	iconType,
+	tooltip,
+	disabled = false,
+	onClick,
+	size = undefined,
```
MajorMeerkatThe3rd, line 20: @LemToUp Here please follow the convention we used for other components (like the AppBuilderTextWidgetComponent, as an example). There should be style props, default props, etc..

Like this the input to this component could probably be much less as all of these icons share the same size, color, colorDisabled, etc...

Additionally, this can then also be edited via the them file.

components/shapediver/viewport/buttons/ViewportIconButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+	iconStyle = IconProps.style,
+}: ViewportIconButtonProps) {
+	return (
+		<TooltipWrapper label={tooltip ?? ""}>
+			<ActionIcon
+				onClick={onClick}
+				disabled={disabled}
+				size={size}
+				variant={disabled ? variantDisabled : variant}
+				aria-label={tooltip ?? "Button"}
```
MajorMeerkatThe3rd, line 34: @LemToUp If there is not tooltip, just leave this empty

components/shapediver/viewport/buttons/ViewportIconButtonDropdown.tsx
```diff
@@ -0,0 +1,85 @@
...
+	visible?: boolean;
+}
+
+export default function ViewportIconButtonDropdown({
+	iconType,
+	tooltip,
+	disabled = false,
+	sections,
+	menuDropdownProps = {style: ViewportTransparentBackgroundStyle},
+	visible = true,
```
MajorMeerkatThe3rd, line 27: @LemToUp Same as above for the ViewportIconButton component, please add themable properties

components/shapediver/viewport/buttons/ViewportIconButtonDropdown.tsx
```diff
@@ -0,0 +1,85 @@
...
+		visible,
+		onClickOutside,
+	);
+
+	const hasItems = sections.some((s) => s.length > 0);
+	if (!hasItems) return null;
+
+	return (
+		<Menu
+			shadow="md"
```
MajorMeerkatThe3rd, line 46: @LemToUp please add this to the theme as well and look for other hard-coded styles where this could be the case

components/shapediver/viewport/buttons/ViewportIconButtonDropdown.tsx
```diff
@@ -0,0 +1,85 @@
...
+			position="bottom-end"
+			opened={visible && isMenuOpened}
+			onChange={setIsMenuOpened}
+		>
+			<Menu.Target>
+				<div>
+					<ViewportIconButton
+						iconType={iconType}
+						tooltip={tooltip ?? "Menu"}
```
MajorMeerkatThe3rd, line 55: @LemToUp As above, please just leave this empty if no tooltip is supplied

store/useShapeDiverStoreViewport.ts
```diff
@@ -41,6 +43,23 @@ export const useShapeDiverStoreViewport = create<IShapeDiverStoreViewport>()(
 
 				try {
 					viewport = await createViewport(dto);
+
+					// Add viewport icons using external store access
+					const iconStore =
+						useShapeDiverViewportIconsStore.getState();
+					iconStore.add(viewport.id, [
+						...(viewport.enableAR
```
MajorMeerkatThe3rd, line 51: @LemToUp Can this enableAR be moved to another point where we check for all the other icons if they can be enabled?

store/useShapeDiverStoreViewport.ts
```diff
@@ -41,6 +43,23 @@ export const useShapeDiverStoreViewport = create<IShapeDiverStoreViewport>()(
 
 				try {
 					viewport = await createViewport(dto);
+
+					// Add viewport icons using external store access
```
MajorMeerkatThe3rd, line 47: @LemToUp this should be moved somewhere else. You could add it to the store you have just created and add a subscribe function to the useShapeDiverStoreViewport to be aware whenever a viewport as been created or removed.

types/store/shapediverStoreViewportIcons.ts
```diff
@@ -0,0 +1,52 @@
...
+	Ar = "ar",
+	Zoom = "zoom",
+	Fullscreen = "fullscreen",
+	Cameras = "cameras",
+	Undo = "undo",
+	Redo = "redo",
+	HistoryMenu = "historyMenu",
+}
+
+export interface ViewportIconButtonDef {
```
MajorMeerkatThe3rd, line 11: @LemToUp Why is this type needed? Shouldn't this be just the ViewportIconButton component type?

types/store/shapediverStoreViewportIcons.ts
```diff
@@ -0,0 +1,52 @@
+export enum ViewportIconButtonType {
```
MajorMeerkatThe3rd, line 1: @LemToUp see my comment below, this shouldn't be needed to be typed

components/shapediver/viewport/ViewportIcons.tsx
```diff
@@ -69,6 +62,91 @@ const defaultStyleProps: ViewportIconsOptionalProps = {
...
+		namespace,
+		buttonsDisabled,
+		executing,
+		hasPendingChanges,
+		iconsVisible,
+		fullscreenId,
+		...commonProps
+	} = context;
+
+	switch (kind) {
```
MajorMeerkatThe3rd, line 90: @LemToUp please already store the component itself in the store, then we don't even have to know anything about the type of viewport here. This should all be abstracted.

components/shapediver/viewport/ViewportIcons.tsx
```diff
@@ -247,124 +251,105 @@ export default function ViewportIcons(
...
+			colorDisabled,
+			variant,
+			variantDisabled,
+			size,
+			iconStyle,
 		],
 	);
 
+	// Render layout items dynamically
+	const dynamicContent = useMemo(() => {
```
MajorMeerkatThe3rd, line 299: @LemToUp I'm not sure what's happening here, why is this needed? Would this be removed either way with my comment from above?

components/shapediver/viewport/ViewportIcons.tsx
```diff
@@ -247,124 +251,105 @@ export default function ViewportIcons(
...
+						);
+						if (button) groupButtons.push(button);
+					});
+					// Add divider between sections within a group
+					if (
+						groupButtons.length > 0 &&
+						section !== item.sections[item.sections.length - 1]
+					) {
+						groupButtons.push(
+							<Divider
```
MajorMeerkatThe3rd, line 325: @LemToUp This logic should happen in the store for the icons. If you already get the complete component there, then it should be simplified quite a bit

types/store/shapediverStoreViewportIcons.ts
```diff
@@ -0,0 +1,52 @@
...
+	layout: ViewportIconLayoutItem[];
+}
+
+export type ViewportIconsStateByViewport = Record<
+	string,
+	ViewportIconViewportState
+>;
+
+export interface IShapeDiverViewportIconsStore {
+	viewports: ViewportIconsStateByViewport;
```
MajorMeerkatThe3rd, line 38: @LemToUp Minor detail, please rename this to viewportIcons, otherwise it get's a bit confusing

MajorMeerkatThe3rd (changes requested):
@LemToUp I left a few comments. 
The main change I would request is, that the store already gets the full components. Like that we don't need a lot of the types and the logic to create them. 
I also requested some changes for the theming of this.
Please see the corresponding comments I left, thank you!

components/shapediver/viewport/buttons/ViewportIconButtonDropdown.tsx
LemToUp: https://github.com/shapediver/AppBuilderShared/commit/40e416aa8c26a9cc0f8005dd8b6ec51fa6cc0802

store/useShapeDiverStoreViewport.ts
LemToUp: @MajorMeerkatThe3rd we don't check it yet:
https://github.com/shapediver/AppBuilderShared/blob/40e416aa8c26a9cc0f8005dd8b6ec51fa6cc0802/hooks/ui/useCustomTheme.ts#L1172

types/store/shapediverStoreViewportIcons.ts
LemToUp: Done: 
https://github.com/shapediver/AppBuilderShared/pull/174/files#diff-2b2c25b1bc102d2132ef26f5a5af4faade2abb07427d87cba0f5f1f23289efcdR1

store/useShapeDiverStoreViewport.ts
LemToUp: Done:
https://github.com/shapediver/AppBuilderShared/pull/174/commits/361784488bdb8f6e8d9023709c76b9c0a119a2aa

components/shapediver/viewport/ViewportIcons.tsx
LemToUp: Related with https://github.com/shapediver/AppBuilderShared/pull/174#discussion_r2284430286

MajorMeerkatThe3rd: See above

components/shapediver/viewport/ViewportIcons.tsx
MajorMeerkatThe3rd: @LemToUp In some stores we already store JSX.Element types. Whats important here is that this whole creation process should be happening in each separate button already. In this file, the logic should be completely abstract. Otherwise, we would have to change this everytime we add new buttons.

MajorMeerkatThe3rd (changes requested):
@LemToUp I think we are almost there. Just another layer of abstraction is needed here. 
As you can see from the comment, no other component should even know that the ArButton (for example) exists, besides the ArButton itself and the component that needs it.

PR #176, opened 2025-08-25
SS-8346 parameter value sources

snabela:
https://shapediver.atlassian.net/browse/SS-8346

### DIFF
--- a/types/shapediver/appbuilder.ts
+++ b/types/shapediver/appbuilder.ts
@@ -11,6 +11,11 @@ import {SessionCreateDto} from "@AppBuilderShared/types/store/shapediverStoreSes
 import {MantineColor} from "@mantine/core";
 import {Gradient} from "@shapediver/viewer.features.attribute-visualization";
 import {TAG3D_JUSTIFICATION} from "@shapediver/viewer.session";
+import {
+	ICameraApi,
+	IOrthographicCameraApi,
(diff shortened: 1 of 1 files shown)


types/shapediver/appbuilder.ts
```diff
@@ -182,6 +187,86 @@ export interface IAppBuilderImageRef {
...
+	resolution?: unknown;
+	/**
+	 * Optional camera settings to be used for the screenshot. Defaults to the current camera of the viewport.
+	 * If a "name" is provided, the settings of the camera with that name are used as a base.
+	 */
+	camera?:
+		| Pick<ICameraApi, "name" | "position" | "target" | "type">
+		| Pick<IPerspectiveCameraApi, "fov">
+		| Pick<IOrthographicCameraApi, "direction">;
+}
```
snabela, line 222: @MajorMeerkatThe3rd to be discussed

snabela: @mathieu-shapediver the idea about `type` was to allow to define the type of camera to be used, without having to specify a `name`. Does that make sense?

mathieu-shapediver: @snabela Our UI requires both a name and a type when creating a camera:
image
Therefore the check should not be necessary, as long as name is unique.

PR #182, opened 2025-09-05
SS-8956: Add NumberInput value validation for non acceptRejectMode

LemToUp:
@snabela Non acceptRejectMode requires different behavior.
https://shapediver.atlassian.net/browse/SS-8956?focusedCommentId=46551


snabela (approved):
Thank you @LemToUp !

PR #184, opened 2025-09-07
SS-8956 alternative approach to number validation

snabela:


snabela:
@LemToUp I started experimenting with this, and found an alternative solution, which uses a separate state for the NumberInput component. I think this will improve the UX nicely, but feedback by @mathieu-shapediver is required. 

@mathieu-shapediver you can try your test example here: 
https://dev-wwwcdn.us-east-1.shapediver.com/app/builder/v1/main/SS-8956---2/?slug=250722-numberinputs-2&redirect=0

The behavior is quite different than it was before: 
  * you can enter numbers out of the range
  * try doing that, and then press enter or tab
  * also try pressing ESC

PR #194, opened 2025-09-15
SS-8949: Hide tab titles and change settings hideExports and hideData…

LemToUp:
https://shapediver.atlassian.net/browse/SS-8949

### DIFF
--- a/components/ui/TabsComponent.tsx
+++ b/components/ui/TabsComponent.tsx
@@ -4,16 +4,16 @@ import {BoxProps, Stack, Tabs} from "@mantine/core";
 import React, {ReactElement, useEffect, useRef, useState} from "react";
 
 interface PropsTab extends BoxProps {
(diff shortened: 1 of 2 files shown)


components/ui/TabsComponent.tsx
snabela, line 1: @LemToUp in case we pass an empty `name`, would there be an error? I am asking whether we need the `hideName` property. I understand that we use the `name` for setting the `value` of the tabs. An alternative solution might be to introduce an optional `value` property, and also make the `name` attribute optional. In case `value` is not given but `name` is, we keep using the `name` as value. Otherwise, we can create random values.

snabela: @LemToUp in that case we could just create some random values

snabela (approved):
Many thanks @LemToUp !

PR #197, opened 2025-09-16
SS-9007: Improve icons rendering logic

LemToUp:
https://shapediver.atlassian.net/browse/SS-9007


snabela:
> @LemToUp looks good to me, I'm just not sure about the render function and it's location. How do we handle this for other context like this? @snabela FYI

I will review this tomorrow. It seems to me that we could move it to the component context, no?

LemToUp:
> @LemToUp @MajorMeerkatThe3rd there are a few things that appear strange to me here:
> 
> * `types/components/shapediver/componentTypes.ts` should only contain types, no implementation
> * It seems to me that the "render" function could nicely be replaced by making use of the component context, please search for `useContext(ComponentContext)` in the codebase

@snabela  @MajorMeerkatThe3rd one more try:
https://github.com/shapediver/AppBuilderShared/pull/197/commits/e17218660ef227f4401f0bbbe36c175eb9caa2c7

MajorMeerkatThe3rd (changes requested):
@LemToUp, just copying yesterdays comments from slack here to have a shared source of information:

Ok I think there was some confusion :sweat_smile:

- I meant that you can look at the parameters component context as an example, but I'm not sure why you moved them now
- my idea was that you can replace the renderButtonByKind by something similar as was already used for parameters and exports
- why has the renderViewportIcons to be a separate function and cannot be withing the useMemo
- why is the React.createElement needed in renderViewportIcons ? You could just create the element there

MajorMeerkatThe3rd (changes requested):
@LemToUp added some comments

PR #199, opened 2025-09-17
SS-8955: return data from instances

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8955

pretty straightforward actually. I added the logic to process the outputs if there are any output actions defined. I took the opportunity to also separate the useAppBuilderInstances hook main useEffect into separate functions to keep this a bit more structured.

### DIFF
--- a/hooks/shapediver/appbuilder/useAppBuilderInstances.ts
+++ b/hooks/shapediver/appbuilder/useAppBuilderInstances.ts
@@ -529,6 +529,7 @@ const processOutputActions = (
 	// to only call the batch update once
 	// this is necessary as multiple instances might want to set the same parameter
 	const outputReturns: {
+		// the parameter id
 		[key: string]: {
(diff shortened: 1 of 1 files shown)


hooks/shapediver/appbuilder/useAppBuilderInstances.ts
```diff
@@ -182,204 +203,417 @@ export function useAppBuilderInstances(props: Props) {
...
+	namespace: string,
+	batchParameterValueUpdate: (params: Record<string, any>) => Promise<void>,
+) => {
+	let outputCallbackPromises: Promise<void>[] = [];
+
+	// we gather all outputs to be set in one object
+	// to only call the batch update once
+	// this is necessary as multiple instances might want to set the same parameter
+	const outputReturns: {
+		[key: string]: {
```
snabela, line 513: @MajorMeerkatThe3rd lowest level is the parameter ID, correct?

PR #201, opened 2025-09-17
SS-8954: option to only create instance sessions on first use

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8954

### DIFF
--- a/types/store/shapediverStoreSession.ts
+++ b/types/store/shapediverStoreSession.ts
@@ -55,8 +55,12 @@ export interface IShapeDiverStoreSession {
 
 	/**
 	 * Create a session and add it to the store.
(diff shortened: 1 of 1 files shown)


types/store/shapediverStoreSession.ts
```diff
@@ -51,6 +62,18 @@ export interface IShapeDiverStoreSession {
 	createSession: (
 		dto: SessionCreateDto,
 		callbacks?: IShapeDiverStoreSessionCallbacks,
+		load?: boolean,
+	) => Promise<ISessionApi | undefined>;
```
snabela, line 66: @MajorMeerkatThe3rd let's document the `load` parameter, and how it is used in combination with `pendingSessions` and `createPendingSession`

PR #209, opened 2025-09-23
SS-8346: parameter value sources

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-8346

@snabela this is now already working as expected for the data outputs. Maybe you have a second to have a look at the implementation. I think I found the most minimal, abstract solution to this!

Next I would look at the screenshot parameter. There I can better test the asynchronous cases


store/useShapeDiverStoreSession.ts
```diff
@@ -420,6 +422,30 @@ const assignOutputUpdateCallback = (
...
+					false,
+				);
+			}
+		} catch (error) {
+			// Silent fail to avoid breaking existing functionality
+			console.warn(
+				`Failed to sync output ${outputApi.id} for session ${sessionApi.id}:`,
+				error,
+			);
+		}
```
snabela, line 448: @MajorMeerkatThe3rd quick question: what could fail (throw) here?

snabela (dismissed):
Great, this will be so powerful, many thanks @MajorMeerkatThe3rd !
I left minor comments.

components/shapediver/appbuilder/actions/AppBuilderActionCameraComponent.tsx
```diff
@@ -0,0 +1,158 @@
...
+	namespace: string;
+};
+
+/**
+ * Functional component for a "camera" action.
+ *
+ * @returns
+ */
+export default function AppBuilderActionCameraComponent(props: Props) {
+	const {label = "Start camera", icon = "tabler:video", tooltip} = props;
```
snabela, line 24: @MajorMeerkatThe3rd we could set the label depending on the type of camera action, i.e. "Reset camera", "Set camera", "Animate camera"

PR #217, opened 2025-10-07
SS-9066: removed hideDataOutputs option

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9066


components/shapediver/appbuilder/AppBuilderFallbackContainerComponent.tsx
```diff
@@ -59,9 +58,7 @@ export default function AppBuilderFallbackContainerComponent({
 			if (hideExportsTab) {
```
mathieu-shapediver, line 58: @MajorMeerkatThe3rd The logic should be the same for exports and outputs, i.e the outputs should be moved to the same codition (hideExportsTab)

MajorMeerkatThe3rd: fixed

PR #218, opened 2025-10-07
SS-9071: implemented model state image support

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9071

@snabela There are some cases that I would like you to have a look at:

1. Whenever a model state is created without any inputs, I now don't include the screenshot. I don't know why we would, do you agree? examples: `AppBuilderActionSetBrowserLocationComponent`, `HistoryMenuButton`, `useKeyBindings` 
2. In the `createModelState` function, the session api is currently directly used for the parameters (https://github.com/shapediver/AppBuilderShared/blob/35b423a8c90c883cd3c17a2dbd7a7687b8c6410d/hooks/shapediver/useCreateModelState.ts#L98) but not for the exports (https://github.com/shapediver/AppBuilderShared/blob/35b423a8c90c883cd3c17a2dbd7a7687b8c6410d/hooks/shapediver/useCreateModelState.ts#L138). I think it makes sense to do it the same way for both, do you agree?
3. I currently just pass through the `href` from the export, is that enough or should I fetch it and then create a dataurl from the blob?

### DIFF
--- a/components/shapediver/appbuilder/actions/AppBuilderActionSetBrowserLocationComponent.tsx
+++ b/components/shapediver/appbuilder/actions/AppBuilderActionSetBrowserLocationComponent.tsx
@@ -64,7 +64,8 @@ export default function AppBuilderActionSetBrowserLocationComponent(
 			const {modelStateId} = await createModelState(
 				undefined, // <-- parameterNamesToInclude: use default according to the theme
 				undefined, // <-- parameterNamesToExclude: use default according to the theme
-				true, // <-- includeImage,
+				false, // <-- includeImage,
+				undefined,
 				undefined, // <-- custom data
 				false, // <-- includeGltf,
 			);
--- a/components/shapediver/viewport/buttons/HistoryMenuButton.tsx
+++ b/components/shapediver/viewport/buttons/HistoryMenuButton.tsx
@@ -34,7 +34,8 @@ export default function HistoryMenuButton({
 		const {modelStateId} = await createModelState(
 			undefined, // <-- parameterNamesToInclude: use default according to the theme
 			undefined, // <-- parameterNamesToExclude: use default according to the theme
-			true, // <-- includeImage,
+			false, // <-- includeImage,
+			undefined, // <-- image
 			undefined, // <-- custom data
 			false, // <-- includeGltf
 		);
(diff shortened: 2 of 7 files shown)


components/shapediver/appbuilder/actions/AppBuilderActionSetBrowserLocationComponent.tsx
```diff
@@ -64,7 +64,8 @@ export default function AppBuilderActionSetBrowserLocationComponent(
 			const {modelStateId} = await createModelState(
 				undefined, // <-- parameterNamesToInclude: use default according to the theme
 				undefined, // <-- parameterNamesToExclude: use default according to the theme
-				true, // <-- includeImage,
+				false, // <-- includeImage,
```
snabela, line 67: @MajorMeerkatThe3rd let's keep `includeImage` set to `true` in this case, at least one user I know is already relying on this.

components/shapediver/viewport/buttons/HistoryMenuButton.tsx
```diff
@@ -34,7 +34,8 @@ export default function HistoryMenuButton({
 		const {modelStateId} = await createModelState(
 			undefined, // <-- parameterNamesToInclude: use default according to the theme
 			undefined, // <-- parameterNamesToExclude: use default according to the theme
-			true, // <-- includeImage,
+			false, // <-- includeImage,
```
snabela, line 37: @MajorMeerkatThe3rd I also suggest to keep `includeImage` set to `true` in this case

hooks/shapediver/useKeyBindings.ts
```diff
@@ -28,7 +28,8 @@ export function useKeyBindings(props: Props) {
 		const {modelStateId, screenshot} = await createModelState(
 			undefined, // <-- parameterNamesToInclude: use default according to the theme
 			undefined, // <-- parameterNamesToExclude: use default according to the theme
-			true, // <-- includeImage,
+			false, // <-- includeImage,
```
snabela, line 31: @MajorMeerkatThe3rd I also suggest to keep `includeImage` set to `true` in this case

snabela (changes requested):
@MajorMeerkatThe3rd many thanks! I answered your questions.

PR #304, opened 2026-04-03
SS-9445: Add Slider min\max override

LemToUp:
https://shapediver.atlassian.net/browse/SS-9445

### DIFF
--- a/entities/parameter/model/useSettingsMinMax.ts
+++ b/entities/parameter/model/useSettingsMinMax.ts
@@ -24,16 +24,22 @@ export function useSettingsMinMax(definition: IShapeDiverParameterDefinition) {
 		? validatedSettings.data.max
 		: undefined;
 
-	const { effectiveMin, effectiveMax } = useMemo(() => {
-		let clampedMin = settingsMin !== undefined ? Math.max(defMin, settingsMin) : defMin;
-		let clampedMax = settingsMax !== undefined ? Math.min(defMax, settingsMax) : defMax;
+	const {effectiveMin, effectiveMax} = useMemo(() => {
+		let clampedMin =
+			settingsMin !== undefined
+				? Math.min(defMax, Math.max(defMin, settingsMin))
+				: defMin;
+		let clampedMax =
+			settingsMax !== undefined
+				? Math.max(defMin, Math.min(defMax, settingsMax))
+				: defMax;
 		if (clampedMin > clampedMax) {
 			[clampedMin, clampedMax] = [clampedMax, clampedMin];
 		}
 		return {
 			effectiveMin: clampedMin,
 			effectiveMax: clampedMax,
-		}
+		};
 	}, [defMin, defMax, settingsMin, settingsMax]);
 
 	useEffect(() => {
@@ -59,7 +65,14 @@ export function useSettingsMinMax(definition: IShapeDiverParameterDefinition) {
 				message: `Min override exceeds max override for "${definition.name}".`,
 			});
 		}
-	}, [settingsMin, settingsMax, defMin, defMax, definition.name, notifications]);
+	}, [
+		settingsMin,
+		settingsMax,
+		defMin,
+		defMax,
+		definition.name,
+		notifications,
+	]);
 
 	return {
 		validatedSettings,
--- a/entities/parameter/model/useSettingsMinMax.ts
+++ b/entities/parameter/model/useSettingsMinMax.ts
@@ -24,10 +24,17 @@ export function useSettingsMinMax(definition: IShapeDiverParameterDefinition) {
 		? validatedSettings.data.max
 		: undefined;
 
-	const { effectiveMin, effectiveMax } = useMemo(() => ({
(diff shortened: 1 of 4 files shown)


entities/parameter/model/useSettingsMinMax.ts
```diff
@@ -0,0 +1,68 @@
...
+				message: `Min override (${settingsMin}) is below the parameter minimum (${defMin}) for "${definition.name}", clamping to parameter minimum.`,
+			});
+		}
+		if (settingsMax !== undefined && settingsMax > defMax) {
+			notifications.error({
+				title: "Invalid Parameter Settings",
+				message: `Max override (${settingsMax}) exceeds the parameter maximum (${defMax}) for "${definition.name}", clamping to parameter maximum.`,
+			});
+		}
+		const clampedMin =
```
MajorMeerkatThe3rd, line 58: I think only the settingsMin, settingsMax should be checked here
for the definition we don't have to, so this can be simplified to

if(settingsMin !== undefined && settingsMax !== undefined && settingsMin > settingsMax)

entities/parameter/model/useSettingsMinMax.ts
```diff
@@ -0,0 +1,68 @@
...
+	const settingsMax = validatedSettings.success
+		? validatedSettings.data.max
+		: undefined;
+
+	const {effectiveMin, effectiveMax} = useMemo(() => {
+		const clampedMin =
+			settingsMin !== undefined ? Math.max(defMin, settingsMin) : defMin;
+		const clampedMax =
+			settingsMax !== undefined ? Math.min(defMax, settingsMax) : defMax;
+		if (clampedMin > clampedMax) {
```
MajorMeerkatThe3rd, line 36: I don't see how this could happen, but better safe than sorry :D

MajorMeerkatThe3rd (changes requested):
@LemToUp I left some very small comments

entities/parameter/model/useSettingsMinMax.ts
LemToUp: @MajorMeerkatThe3rd the idea is - if custom limits break default ones -> use default.

entities/parameter/model/useSettingsMinMax.ts
LemToUp: Good point, simplified - https://github.com/shapediver/AppBuilderShared/pull/304/changes/c249b8908fe61fbd2d53a3041d7e6259d6de0049

entities/parameter/model/useSettingsMinMax.ts
MajorMeerkatThe3rd: @LemToUp but the logic has a general issue. The useMemo doesn't do these checks, so the notifications might be sent but the values are not adjusted accordingly. This is especially an issue for the last check.

Let's say definition is {min: 0, max: 100} and settings are {min: 80, max 20}, then the effective settings would be  {min: 80, max 20} which is incorrect

MajorMeerkatThe3rd (changes requested):

entities/parameter/model/useSettingsMinMax.ts
LemToUp: @MajorMeerkatThe3rd what about another approach ?:
Swap effectiveMin with effectiveMax in case of (min > max) at the upper block.

But we don't need the last warning in that case: "Min override exceeds max override for....."

MajorMeerkatThe3rd: @LemToUp do you mean something like this? That would make sense

```
const { effectiveMin, effectiveMax } = useMemo(() => {
    // Apply clamping first
    const clampedMin = settingsMin !== undefined ? Math.max(defMin, settingsMin) : defMin;
    const clampedMax = settingsMax !== undefined ? Math.min(defMax, settingsMax) : defMax;

    if (clampedMin > clampedMax) {
        // If they cross over, we "ignore" overrides by falling back to the parameter defaults
        return { effectiveMin: defMin, effectiveMax: defMax };
    }

    return { effectiveMin: clampedMin, effectiveMax : clampedMax };
}, [defMin, defMax, settingsMin, settingsMax]);
```

LemToUp: @MajorMeerkatThe3rd exactly + get rid of "Min override exceeds max override for....." message

MajorMeerkatThe3rd: @LemToUp you can still keep the message but only check the settingsMin and settingsMax as I suuggested. Just so that the user knows he made a mistake

LemToUp: @MajorMeerkatThe3rd done:
https://github.com/shapediver/AppBuilderShared/pull/304/changes/4c4547c3002cc29d2599c1c6647315606159dbb7

MajorMeerkatThe3rd (approved):
@LemToUp I added a tiny change just to be sure and am deploying to development

PR #306, opened 2026-04-16
SS-9456: Add default containers

LemToUp:
https://shapediver.atlassian.net/browse/SS-9456?focusedCommentId=67720


snabela (approved):
Thank you @LemToUp !
A related question: Why did this cause a problem for PERI only?

LemToUp:
> Thank you @LemToUp ! A related question: Why did this cause a problem for PERI only?

It used LibraryRoot component without defined container:
https://github.com/shapediver/AppBuilderPERI/blob/846e40f82d6a4a284ab4213f214ae898a11f3364/src/LibraryRoot.tsx#L125

No container - no render

Rest examples usually use AppBuilderRoot

snabela:
> AppBuilderRoot

got it, many thanks for explaining!

PR #307, opened 2026-04-16
Task/ss 9536 export import fix

LemToUp:
https://shapediver.atlassian.net/browse/SS-9536


snabela (approved):
Many thanks @LemToUp !

PR #309, opened 2026-05-07
SS-9298 app builder documentation

LemToUp:
https://shapediver.atlassian.net/browse/SS-9298
related PR: https://github.com/shapediver/AppBuilderSdk/pull/130


entities/parameter/ui/ParameterColorComponent.tsx
```diff
@@ -45,12 +55,14 @@ export function ParameterColorComponentThemeProps(
...
-	props: PropsParameterComponent & Partial<PropsParameterWrapper>,
+	props: PropsParameterComponent &
+		Partial<PropsParameterWrapper> &
+		ParameterColorComponentPropsType,
 ) {
 	const {colorFormat} = useProps(
 		"ParameterColorComponent",
 		defaultStyleProps,
-		defaultStyleProps,
+		props,
```
LemToUp, line 53: Looks like bugfix.

features/appbuilder/ui/AppBuilderActionComponent.tsx
```diff
@@ -24,11 +24,29 @@ type Props = IAppBuilderActionPropsCommon &
...
 
-const defaultStyleProps: Partial<Props> = {
+/**
+ * Theme-driven Mantine `Button` defaults for action buttons.
+ *
+ * @docAttached
+ * @configPath themeOverrides.components.AppBuilderActionComponent.defaultProps
+ * @displayName AppBuilderActionComponent
+ */
+export interface AppBuilderActionComponentStyleProps {
```
LemToUp: Used props extracted for the parser

shared/ui/modal/ModalBase.tsx
```diff
@@ -167,7 +172,7 @@ export default function ModalBase(props: Props & Partial<StyleProps>) {
 		...rest
 	} = props;
 
-	const styleProps = useProps("UniversalModal", defaultStyleProps, rest);
+	const styleProps = useProps("ModalBase", defaultStyleProps, rest);
```
LemToUp, line 182: Renamed to be convenient to the config:
https://github.com/shapediver/AppBuilderShared/blob/development/shared/ui/theme/useCustomTheme.ts#L1018

widgets/appbuilder/ui/AppBuilderTextWidgetComponent.tsx
```diff
@@ -13,11 +13,20 @@ import {
...
 import React, {useContext} from "react";
 
-type StylePros = PaperProps;
+/**
+ * Text / markdown widget wrapped in Mantine `Paper`. Theme `defaultProps` follow `PaperProps`.
+ *
+ * @docAttached
+ * @configPath themeOverrides.components.AppBuilderTextWidgetComponent.defaultProps
+ * @displayName AppBuilderTextWidgetComponent
+ * @docLink https://mantine.dev/core/paper/?t=props
```
LemToUp: @docLink tag used

entities/parameter/ui/ParameterColorComponent.tsx
MajorMeerkatThe3rd: absolutely, good catch

MajorMeerkatThe3rd (approved):
@LemToUp thank you, very clean now

widgets/appbuilder/ui/AppBuilderTextWidgetComponent.tsx
snabela: @LemToUp got it, many thanks for pointing this out!

snabela (approved):
Many thanks @LemToUp !

entities/output/ui/OutputStargateComponent.tsx
```diff
@@ -1,37 +1,42 @@
...
-	accordionProps?: AccordionProps;
-	accordionItemProps?: AccordionItemProps;
-	accordionControlProps?: AccordionControlProps;
-	accordionPanelProps?: AccordionPanelProps;
+/**
+ * @docAttached
+ * @configPath themeOverrides.components.OutputStargateComponent.defaultProps
+ * @displayName OutputStargateComponent
+ */
+export interface OutputStargateComponentStyleProps {
```
MajorMeerkatThe3rd, line 25: @LemToUp are the props here automatically documented or is this missing?
not only an issue for this object but in general

MajorMeerkatThe3rd (changes requested):
@LemToUp generally I think this is already pretty great. One question for me is the documentation of properties in general. Are these properties documented automatically? Or at a different place?

Think of it from a use perspective. When I want to change a style property I want to know that that property actually does. But maybe I just overlooked this!

entities/output/ui/OutputStargateComponent.tsx
LemToUp: @MajorMeerkatThe3rd yes, it tries to take from (by priority):

1. predefined list (mantine generics), 
2. describe existing type then,
3. if existing type is not basic (function, library interface columns) - may be artifact generated

https://github.com/shapediver/AppBuilderSdk/blob/0ef79f0ade152fc6df3218660d19a0749dfd9da6/public/doc-flat.json#L1274

But, I don't believe that we can reach the ideal state. Too many nuances and type inherit layers.

MajorMeerkatThe3rd: @LemToUp got it, thank you!

MajorMeerkatThe3rd (approved):

PR #310, opened 2026-05-13
SS-9456: Complete FSD migration in PERI and IJewel3D App Builder repos

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9456

I had to do some general changes as the ijewel3d setup won't work with the index files. Files were imported that shouldn't be


snabela (commented):
Oh wow, that's quite a change. Do you have more details on the problem in the iJewel3D main repo causing this? Maybe this is about settings of vite or the TypeScript compiler?

snabela (approved):

PR #312, opened 2026-05-21
SS-9463 type theme properties, document them

LemToUp:
https://shapediver.atlassian.net/browse/SS-9463
related PR: https://github.com/shapediver/AppBuilderSdk/pull/133


entities/export/config/ExportLabelComponent.types.ts
```diff
@@ -0,0 +1,10 @@
+import {z} from "zod";
+
+/** Theme `defaultProps` for `useProps("ExportLabelComponent", …)`. */
+export const ExportLabelComponentThemeDefaultPropsSchema = z.strictObject({
+	fontWeight: z.string().optional(),
```
MajorMeerkatThe3rd, line 5: @LemToUp Would it be possible to extend this so that other label properties can be exposed as well? I'm thinking about the default Label properties from Mantine, maybe Mantine already exposes something like that.

This is a general topic, not just for the Label, but for all other Mantine components as well.

entities/parameter/config/theme/parameterInteractionThemeDefaultProps.ts
```diff
@@ -0,0 +1,50 @@
+import {z} from "zod";
+
+/**
+ * JSON-serializable values accepted in theme `defaultProps` for fields typed as
+ * `InteractionEffect` in parameter interaction themes (before `parseInteractionEffect`).
+ */
+export const InteractionEffectThemeJsonSchema = z.union([
```
MajorMeerkatThe3rd: @LemToUp all of these properties are already exposed by the viewer, please check and let me know if you have any issues

entities/parameter/model/interaction/useDragManager.ts
```diff
@@ -1,3 +1,4 @@
+import {setOccludeBySceneGeometry} from "@AppBuilderLib/entities/parameter/config/interactionParameterProps.augment";
```
MajorMeerkatThe3rd: @LemToUp I don't see the corresponding file in this PR

features/appbuilder/config/themeComponentDefaultPropsRegistry.ts
MajorMeerkatThe3rd, line 1: @LemToUp in this file we just need to be careful regarding other Repos. For example in the ijewel repo we have to be careful that the interaction components are not imported as some of the viewer packages are not used there.

pages/config/AppBuilderHorizontalContainer.types.ts
```diff
@@ -0,0 +1,30 @@
+import {z} from "zod";
+
+/** JSON theme validation for `AppBuilderHorizontalContainer` (`Group`). `justify` stays JSON-wide (`z.string`); `wrap` matches Mantine `FlexWrap`. */
+const mantineSpacingJsonSchema = z.union([
```
MajorMeerkatThe3rd: @LemToUp is this something that can be directly exposed from Mantine somehow? I'm worried that if there are changes on their end that we have issues

MajorMeerkatThe3rd (commented):
@LemToUp I left a few comments, I think in general this is a great start. We should look into if Mantine components already expose type checks, and if not, how we could do that ourselves. I think the goal should be that users can also change general properties of Mantine components

entities/export/config/ExportLabelComponent.types.ts
LemToUp: @MajorMeerkatThe3rd it focuses on 'useProps'
https://github.com/shapediver/AppBuilderShared/blob/3c151e28ba4a0637526414505b6ea2f4823e71a1/entities/export/ui/ExportLabelComponent.tsx#L48

entities/parameter/config/theme/parameterInteractionThemeDefaultProps.ts
LemToUp: Done: https://github.com/shapediver/AppBuilderShared/commit/26a35aea8f83506dadb7c195d7d99b324425bc1c

MajorMeerkatThe3rd (dismissed):
@LemToUp from my side this looks great now
@snabela maybe you want to have a look as well

shared/mantine-props/text.zod.ts
```diff
@@ -0,0 +1,10 @@
+// Generated by ts-to-zod
+import { z } from "zod";
+
+export const mantineSpacingSchema = z.union([z.literal("xs"), z.literal("sm"), z.literal("md"), z.literal("lg"), z.literal("xl"), z.string(), z.number()]);
```
snabela: @LemToUp I noticed that this type exists multiple times. It doesn't really matter, but I was just wondering whether this is a result from autogeneration?

snabela (approved):
Great, many thanks @LemToUp !

PR #319, opened 2026-06-09
SS-9734: fixed accept/reject race condition

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9734

This time for real! The issue was a race condition where sometimes the parameters were registered before the `acceptRejectMode` was properly set. This lead to the `acceptRejectMode` being undefined.


snabela (dismissed):
Got it, many thanks for digging into this @MajorMeerkatThe3rd !

PR #320, opened 2026-06-12
SS-9153: implemented settings overrides logic for parameters with existing settings

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9153

pretty straightforward, see Jira task for the model I tested this with

### DIFF
--- a/lib/mergeOverrides.ts
+++ b/lib/mergeOverrides.ts
@@ -1,6 +1,7 @@
 /**
  * Deep-merge `source` into `target`.
  * Objects are merged recursively. Arrays and primitives in `source` replace those in `target`.
+ * If a property in `source` is `null`, it is deleted from the result rather than set to `null`.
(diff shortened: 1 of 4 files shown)


lib/mergeOverrides.ts
```diff
@@ -0,0 +1,69 @@
...
+            result[k] = deepMerge(
+                tv as Record<string, unknown>,
+                sv as Record<string, unknown>,
+            );
+        } else {
+            result[k] = sv;
+        }
+    }
+    return result;
+}
```
snabela, line 36: @MajorMeerkatThe3rd would it make sense to support "unsetting" a property (of the overridden value is null)?

MajorMeerkatThe3rd: @snabela I've implemented this now, makes sense! please re-review

snabela (approved):

PR #321, opened 2026-06-15
SS-9756: fixed race condition issue between multiple use effects

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9756

I had to switch some parts to be done synchronously, as otherwise we got into these race conditions.
This was specifically a case for models that waited for output actions from instances


snabela (approved):
thank you @MajorMeerkatThe3rd !

PR #323, opened 2026-06-17
SS-9698: Add stack case: floated container

LemToUp:
https://shapediver.atlassian.net/browse/SS-9698


MajorMeerkatThe3rd (approved):
@LemToUp thank you

PR #328, opened 2026-06-23
SS-9800: fix undo button for custom parameters

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9800

When using undo/redo on a model with instances, custom parameter values (material selections, vase sizes) were reset to their output defaults instead of being restored to the correct previous state. History restore was restoring all namespaces in one parallel batch, including `*_appbuilder` dynamic custom-parameter namespaces. But these namespaces are derived from the AppBuilder output and their parameter stores only exist after the model customization completes.

History undo/redo now follows the normal parameter update path:

1. Restore only primary/model namespaces: *_appbuilder namespaces are excluded from direct restore since they are derived state, not source state.
2. Repopulate derived custom parameters from the restored input: When the AppBuilder output recreates dynamic custom parameters after a model update (including undo), the hook now reads the restored values from the AppBuilder input and uses them instead of falling back to fresh output defaults.

### DIFF
--- a/entities/parameter/model/useShapeDiverStoreParameters.ts
+++ b/entities/parameter/model/useShapeDiverStoreParameters.ts
@@ -6,6 +6,7 @@ import {
 	IShapeDiverOutput,
 	IShapeDiverOutputDefinition,
 } from "@AppBuilderLib/entities/output/config/output";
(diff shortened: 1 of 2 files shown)


snabela (approved):
Makes complete sense, thank you for digging into this @MajorMeerkatThe3rd !

PR #331, opened 2026-06-25
SS-9800: implemented context aware parameter history

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9800

### What changed                                                                           
                                                                                            
 - During history restore, only real/controller session namespaces are executed             
 - *_appbuilder history state is kept temporarily and used only to rehydrate the derived    
   custom parameter stores                                                                  
 - Rehydration updates the custom parameter UI/store values directly, without executing     
   their executor again                                                                     
                                                                                            
 ### Why                                                                                    
                                                                                            
 History stores a snapshot of resulting state, not the original single user action.         
 So undo/redo should restore the source-of-truth session state and then rehydrate derived   
 AppBuilder UI state, instead of rerunning the full live custom-parameter execution path.

### DIFF
--- a/entities/parameter/config/shapediverStoreParameters.ts
+++ b/entities/parameter/config/shapediverStoreParameters.ts
@@ -299,6 +299,13 @@ export interface IShapeDiverStoreParameters {
 	readonly historyIndex: number;
 
 	/**
+	 * Transient derived namespace state captured during history restore.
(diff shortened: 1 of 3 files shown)


snabela (commented):
Danke @MajorMeerkatThe3rd !

PR #332, opened 2026-06-29
SS-9823: mantine v9

LemToUp:
https://shapediver.atlassian.net/browse/SS-9823

@snabela do we need legacy colors scheme ?:
https://mantine.dev/guides/8x-to-9x/#light-variant-color-changes


snabela (approved):
Thank you @LemToUp !

PR #336, opened 2026-07-16
SS-9464: New App Builder toolbars

MajorMeerkatThe3rd:
https://shapediver.atlassian.net/browse/SS-9464

This PR adds a new toolbar system to App Builder. Toolbars can now be configured directly in App Builder JSON and used to promote important viewer, configurator, export, camera, and model-state features around the viewport. The default viewport toolbar has also been moved to this new system, so custom toolbars and the built-in toolbar now work the same way.

### Add configurable toolbars around the viewer

App Builder can now place toolbars in viewport slots. A toolbar chooses:

- a viewport edge: `top`, `right`, `bottom`, or `left`
- an alignment on that edge: `start`, `center`, or `end`
- an order, so multiple toolbars in the same slot can stack predictably
- a visibility mode, such as always visible or shown based on mouse activity

### Use actions and controls as toolbar items

Toolbar items can be either direct triggers or popover triggers.

Direct triggers run immediately when clicked. Examples:

- action items execute an App Builder action, such as zoom, fullscreen, reset parameters, set parameter values, add to cart, or play sound
- export refs trigger the referenced export as a toolbar export control

Popover triggers open additional UI. Examples:

- parameter refs open the referenced parameter in a popover
- output refs open the referenced output in a popover
- menu items open action menus
- widget items open widget panels
- tab items open tabbed panels

This keeps the toolbar flexible: simple operations can be one-click buttons, while more complex workflows can live behind a compact popover. Popover styling can be adjusted and still supports the legacy viewport-icon theme fallback.

Toolbar popovers are created by using toolbar items that contain nested content instead of a direct action.

For example:

- `items` creates a menu popover
- `widgets` creates a widget panel popover
- `tabs` creates a tabbed popover
- `type: "parameter"` creates a parameter popover
- `type: "output"` creates an output popover

Menus can be flat or grouped. Grouped menus render dividers automatically between sections. This makes it easy to create compact utility menus without custom React code.

### New Actions

Existing App Builder actions can be used in toolbar buttons and toolbar menus, including camera actions, set parameter values, add to cart, close configurator, message parent, set browser location, and sound actions.
                                                                                                                                  
New action types added as part of this work include: 

- AR
- Import parameter values
- Export parameter values
- Fullscreen / fullscreen 3-state
- Import model state
- Redo
- Undo
- Reset parameter values

## Legacy compatibility

The old `ViewportIcons` rendering path has been removed from active App Builder usage.

Legacy theme keys are still accepted for compatibility:

- `ViewportIcons`
- `ViewportIconButton`

They now map into the new toolbar system instead of rendering the old components.

New preferred theme keys are:

- `AppBuilderToolbar`
- `AppBuilderToolbarLayer`
- `AppBuilderToolbarButton`
- `AppBuilderToolbarIconButton`
- `AppBuilderToolbarMenuItemButton`

@snabela please focus on the `appbuilder.ts` to see if the general structure is to your liking
@LemToUp please review in general


features/appbuilder/config/appbuilder.ts
```diff
@@ -1151,16 +1251,89 @@ export type AppBuilderAnchor3dContainerProperties = {
...
+	IAppBuilderControl;
+
+export type IAppBuilderToolbarActionItem = IAppBuilderToolbarItemBase & {
+	type: "action";
+	props: IAppBuilderControlActionRef;
+};
+
+export type IAppBuilderToolbarMenuItem = IAppBuilderToolbarItemBase & {
+	/** Additional toolbar actions shown when this item is opened. */
+	items: IAppBuilderToolbarActionItem[] | IAppBuilderToolbarActionItem[][];
```
snabela: @MajorMeerkatThe3rd if I get it right, `IAppBuilderToolbarActionItem[][]` would be groups of action items?

features/appbuilder/config/appbuilder.ts
```diff
@@ -1151,16 +1251,89 @@ export type AppBuilderAnchor3dContainerProperties = {
...
+	presentation?: "button" | "item";
+}
+
+export type IAppBuilderToolbarControlItem = IAppBuilderToolbarItemBase &
+	IAppBuilderControl;
+
+export type IAppBuilderToolbarActionItem = IAppBuilderToolbarItemBase & {
+	type: "action";
+	props: IAppBuilderControlActionRef;
+};
```
snabela: @MajorMeerkatThe3rd what do you think of adding `type` and `props` to `IAppBuilderToolbarItemBase`, always use `props` fur further properties, and discriminating by `type`?

snabela (commented):
@MajorMeerkatThe3rd great work, very much looking forward to this! 

I left some comments regarding the types, probably we should have a quick call.

features/appbuilder/config/appbuilder.ts
MajorMeerkatThe3rd: @snabela correct! It's used for the nesting that is then divided by separators, you can see it in this example: https://github.com/shapediver/AppBuilderSdk/blob/b9f65b2a72dc3a7fcfdb73de7114623fb20e60c6/public/example-toolbar-container.json

features/appbuilder/config/appbuilder.ts
MajorMeerkatThe3rd: @snabela makes sense! I'm not doing it now as other things might come up as well and I'll just do everything at once

PR #340, opened 2026-07-24
SS-9817: Make buttonFlex configurable

LemToUp:
https://shapediver.atlassian.net/browse/SS-9817

@snabela it will allows to switch button flex direction to vertical for small screens via theme file and do not replace text by icons.


snabela (approved):
Nice! Many thanks @LemToUp !

PR #344, opened 2026-07-29
SS-9721: Add VITE_STATE_PROTECTION .env variable to manage state prot…

LemToUp:
…ection

https://shapediver.atlassian.net/browse/SS-9721?focusedCommentId=80676


snabela (approved):
Many thanks @LemToUp !

PR #345, opened 2026-07-30
SS-9721: Add stateProtection theme property

LemToUp:


MajorMeerkatThe3rd (approved):
@LemToUp great, thank you

PR #347, opened 2026-07-31
SS-9940: Fix redirect=0 if saved state exists

LemToUp:
https://shapediver.atlassian.net/browse/SS-9940


MajorMeerkatThe3rd (approved):
@LemToUp thank you

snabela (commented):
@LemToUp @MajorMeerkatThe3rd this solution does not seem great to me, because it uses a second (unnecessary) call to 

Please see here: https://github.com/shapediver/AppBuilderShared/blob/ef08cb3a65ee4b64e0e4190d7d1b3826dcad5ef3/features/appbuilder/model/useResolveAppBuilderSessions.ts#L166
We already store the saved states that come along with the iframe API call to the saved state store. We should use the data from this store (useShapeDiverStorePlatformSavedStates)

PR #352, opened 2026-08-07
SS-9973 Schema definition for agents

snabela:
https://shapediver.atlassian.net/browse/SS-9973

### DIFF
--- a/features/appbuilder/config/appbuilder.ts
+++ b/features/appbuilder/config/appbuilder.ts
@@ -14,6 +14,7 @@ import {
 	OrthographicCameraProperties,
 	PerspectiveCameraProperties,
 } from "@shapediver/viewer.shared.types";
+import {IAppBuilderAgent} from "./appbuilderagent";
 import {
 	IAppBuilderWidgetPropsAreaChart,
 	IAppBuilderWidgetPropsBarChart,
@@ -375,6 +376,8 @@ export interface IAppBuilderActionDefinition {
 
 /** Common properties of App Builder action controls and legacy actions. */
 export interface IAppBuilderActionPropsCommon {
+	/** Optional identifier of the action. Used to uniquely reference actions in agent definitions, etc. */
+	id?: string;
 	/** Label (of the button etc). Optional, defaults to a value depending on the type of action. Set to empty string to show only an icon. */
 	label?: string;
 	/** Optional icon (of the button etc). */
@@ -404,7 +407,8 @@ export type AppBuilderParameterValueSourceType =
 	| "dataOutput"
 	| "export"
 	| "sdtf"
-	| "modelState";
+	| "modelState"
+	| "agentTool";
 
 /**
  * Properties for the "screenshot" parameter value source.
@@ -508,6 +512,18 @@ export interface IAppBuilderParameterValueSourcePropsModelState extends IAppBuil
 	updateUrl?: boolean;
 }
 
+/**
+ * Properties for the "agentTool" parameter value source.
+ * This source is used for actions that are triggered by an agent tool.
+ */
+export interface IAppBuilderParameterValueSourcePropsAgentTool {
+	/**
+	 * JSON path to the value in the agent tool's input data that should be used as the parameter value.
+	 * @see https://www.rfc-editor.org/info/rfc9535/
+	 */
+	jsonPath: string;
+}
+
 /** Definition of a parameter value source. */
 export interface IAppBuilderParameterValueSourceDefinition {
 	/** Type of the parameter value source. */
@@ -518,7 +534,8 @@ export interface IAppBuilderParameterValueSourceDefinition {
 		| IAppBuilderParameterValueSourcePropsDataOutput
 		| IAppBuilderParameterValueSourcePropsExport
 		| IAppBuilderParameterValueSourcePropsSdtf
-		| IAppBuilderParameterValueSourcePropsModelState;
+		| IAppBuilderParameterValueSourcePropsModelState
+		| IAppBuilderParameterValueSourcePropsAgentTool;
 }
 
 /** Type used for parameter value definitions */
@@ -1499,6 +1516,16 @@ export interface IAppBuilder {
 	 * Instances are used to customize a session by setting parameters and transformations.
 	 */
 	instances?: IAppBuilderInstanceDefinition[];
+
+	/**
+	 * Optional list of agents.
+	 * For now only one agent is supported, but in the future we might support
+	 * multiple agents, or allow agents to reference others as sub-agents.
+	 * We do not consider parametric updates to the agent definition, i.e.,
+	 * the agents get initialized on loading of the app, and updates to this
+	 * property due to parameter changes are ignored.
+	 */
+	agents?: IAppBuilderAgent[];
 }
 
 /** assert default containers */
--- a/features/appbuilder/config/appbuilderagent.ts
+++ b/features/appbuilder/config/appbuilderagent.ts
@@ -8,6 +8,7 @@ import {
 /** Names of generic tool definitions */
 export type GenericToolName =
 	| "list_parameter_definitions"
+	| "get_parameter_values"
 	| "set_parameter_values"
 	| "list_action_controls"
 	| "trigger_action_control"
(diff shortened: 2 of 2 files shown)


features/appbuilder/config/appbuilder.ts
```diff
@@ -1499,6 +1516,16 @@ export interface IAppBuilder {
...
+	/**
+	 * Optional list of agents.
+	 * For now only one agent is supported, but in the future we might support
+	 * multiple agents, or allow agents to reference others as sub-agents.
+	 * We do not consider parametric updates to the agent definition, i.e.,
+	 * the agents get initialized on loading of the app, and updates to this
+	 * property due to parameter changes are ignored.
+	 */
+	agents?: IAppBuilderAgent[];
 }
```
snabela, line 1529: @LemToUp @mathieu-shapediver @MajorMeerkatThe3rd @edwinhernandezgomez please review and comment starting from here. I consider it important for us to have a common understanding before Dmitry gets started with the implementation (implementation will be a separate task).

features/appbuilder/config/appbuilderagent.ts
```diff
@@ -0,0 +1,250 @@
...
+	 * Optional remote execution settings for the tool.
+	 * Will be ignored if actionSequence is provided.
+	 */
+	remoteExecution?: RemoteToolExecutionSettings;
+}
+
+/**
+ * Definition of an agent that can be used with App Builder.
+ */
+export interface IAppBuilderAgent {
```
LemToUp, line 229: @snabela Will it be some kind of instructions that defines agent expertize for agent initialization ?

features/appbuilder/config/appbuilder.ts
MajorMeerkatThe3rd: @snabela I'm not sure the agent definition at this place makes sense if it's not parametric. I think this could cause issues in the future. Wouldn't it make more sense to have a separate output with a reserved name, like the materialDatabase?

features/appbuilder/config/appbuilderagent.ts
```diff
@@ -0,0 +1,250 @@
...
+
+export type ListParameterDefinitionsToolSettings = {
+	name: "list_parameter_definitions";
+
+	/**
+	 * Optional list of parameters that should be exposed to the agent.
+	 * In case this list is not provided, parameters will be filtered
+	 * based on the @see ListParameterDefinitionsToolSettings.filter property.
+	 */
+	parameters?: IAgentParameterRef[];
```
MajorMeerkatThe3rd, line 70: @snabela Why wouldn't these be our normal parameter definitions? specifying a separate type for it feels redundant. I think a user in AB would just like to connect a parameter control to this

features/appbuilder/config/appbuilderagent.ts
```diff
@@ -0,0 +1,250 @@
...
+
+	/**
+	 * Optional filter for parameters that should be exposed to the agent.
+	 * This property is ignored if the @see ListParameterDefinitionsToolSettings.parameters property is provided.
+	 */
+	filter?: {
```
MajorMeerkatThe3rd, line 76: @snabela Why is this needed when the user specifies the parameters in the first place?

features/appbuilder/config/appbuilderagent.ts
```diff
@@ -0,0 +1,250 @@
...
+	inputSchema: JSONSchema;
+
+	/**
+	 * Optional sequence of actions that should be run when the tool is triggered.
+	 * Information about these actions will not be exposed to the agent.
+	 * Values from the @see SpecificToolSettings.inputSchema can be mapped to the action
+	 * properties using the "agentTool" parameter value source.
+	 * @see IAppBuilderParameterValueSourcePropsAgentTool
+	 */
+	actionSequence?: IAppBuilderActionDefinition[];
```
MajorMeerkatThe3rd, line 217: @snabela I think this skips ahead of this task, right? As actions sequence might be a different object and have some different properties
https://shapediver.atlassian.net/browse/SS-9952

features/appbuilder/config/appbuilderagent.ts
```diff
@@ -0,0 +1,250 @@
...
+
+	/**
+	 * Optional remote execution settings for the tool.
+	 * Will be ignored if actionSequence is provided.
+	 */
+	remoteExecution?: RemoteToolExecutionSettings;
```
MajorMeerkatThe3rd, line 223: @snabela what would this do?

features/appbuilder/config/appbuilder.ts
MajorMeerkatThe3rd: @snabela Also a follow-up question, why wouldn't it parametric?

MajorMeerkatThe3rd (commented):
@snabela I left some comments!

features/appbuilder/config/appbuilderagent.ts
snabela: @MajorMeerkatThe3rd in GH, casting from a parameter reference/control would work (we only need name and optional sessionId). I defined a separate type due to the additional context for the agent, which will be important

features/appbuilder/config/appbuilderagent.ts
snabela: @MajorMeerkatThe3rd by default (if parameters to expose are not defined explicitly), the filter criteria apply.

features/appbuilder/config/appbuilderagent.ts
snabela: @MajorMeerkatThe3rd yes agreed, the actions are still a bit more handwavy. The first implementation task for the agent will not include actions, just parameters.

features/appbuilder/config/appbuilderagent.ts
snabela: For PERI, there will be at least one tool running on their side (hosted by them on a backend). In general, this will allow connecting tools that are not backed by logic in the App Builder frontend (other subagents, remote MCP calls, etc).

features/appbuilder/config/appbuilder.ts
snabela: We don't want the definition of the agent to change during the agent loop. It would just complicate things. It's especially important for prompt caching that the tools and the system prompt don't change (by the way, very good article on prompt caching here: https://earendil.com/posts/prompt-caching/). 

I agree that the App Builder output might not be the best place to define this. I'm still torn about this. For now it's fine to keep it here (and simplifies testing), we can later decide to move it to a separate data output.

features/appbuilder/config/appbuilderagent.ts
snabela: @MajorMeerkatThe3rd I didn't get your question 😅

snabela: oops, wrong recipient 😅 
@LemToUp this interface defines the agent's system prompt and the tools that are available to the agent, depending on the specific use case. Also, there will likely have to be some settings on the chat UI, but I didn't specify this in detail yet.

features/appbuilder/config/appbuilder.ts
MajorMeerkatThe3rd: @snabela While I get that it gets more complicated to make it parametric, don't you see use cases where this could be needed? I'm just worried that if we need this later on, it would be much more difficult than adding it from the start.

snabela: @MajorMeerkatThe3rd I'm not worried about the complication it would add, I just think that the agent definition shouldn't be parametric (the system prompt of the agent, which tools are available to it). Whenever I think of use cases of this, I can see lots of confusion being added. It's like having a subordinate whom you give a task, and whenever he does a step towards the goal, he gets different instructions and the goal might change 😅

features/appbuilder/config/appbuilderagent.ts
LemToUp: @snabela what difference between agents ? Different models or different prompts ?

snabela: @LemToUp the purpose of multiple agents would be to use subagents as tools for the main agent, so each agent would have different prompts, probably different tools, and likely also use different models.

